PR #180, opened 2023-02-20
feat: Endpoint to get QR json offer

x1m3:
Endpoint to get a json  to give instructions to the mobile app.

Closes https://polygonid.atlassian.net/browse/ID-149


internal/api/server.go
```diff
@@ -244,6 +247,67 @@ func (s *Server) GetClaims(ctx context.Context, request GetClaimsRequestObject)
...
+				},
+			},
+			Url: fmt.Sprintf("%s/v1/agent", strings.TrimSuffix(hostURL, "/")),
+		},
+		From: claim.Issuer,
+		Id:   id.String(),
+		Thid: id.String(),
+		To:   *claim.Identifier,
+		Typ:  "application/iden3comm-plain-json",
+		Type: "https://iden3-communication.io/credentials/1.0/offer",
```
javip97, line 307:
```
packers.MediaTypePlainMessage
protocol.CredentialIssuanceResponseMessageType
```

api/api.yaml
```diff
@@ -292,6 +292,31 @@ paths:
           $ref: '#/components/responses/400'
         '500':
           $ref: '#/components/responses/500'
+  /v1/{identifier}/claims/{id}/qrcode:
+    get:
+      summary: Get Claim QR code
+      operationId: GetQrCodeClaim
```
javip97, line 298: I would put `GetClaimQrCode` instead of `GetQrCodeClaim`

javip97: The same with the object response (`GetQrCodeClaimResponse`)

x1m3: Done

internal/api/server.go
x1m3: Done

javip97: my bad, the type should be 

`Type: string(protocol.CredentialOfferMessageType)`

javip97 (approved):
Nice Job!!!

PR #190, opened 2023-03-01
Id 221 ship create api skeleton for admin UI

martinsaporiti:
## Notes
* New admin server ()
* Docker-compose for the admin server
* makefile updated


internal/api_admin/server.go
```diff
@@ -0,0 +1,86 @@
...
+// Server implements StrictServerInterface and holds the implementation of all API controllers
+// This is the glue to the API autogenerated code
+type Server struct {
+	cfg              *config.Configuration
+	identityService  ports.IdentityService
+	claimService     ports.ClaimsService
+	schemaService    ports.SchemaService
+	publisherGateway ports.Publisher
+	packageManager   *iden3comm.PackageManager
+	health           *health.Status
```
x1m3, line 25: Not sure if we are gonna need all of these.  But it is ok.

internal/api_admin/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	switch err.(type) {
+	case AuthError:
+		w.WriteHeader(http.StatusUnauthorized)
+		w.Header().Add("WWW-Authenticate", `Basic realm="restricted", charset="UTF-8"`)
+		_, _ = w.Write([]byte("\"Unauthorized\""))
+	default:
+		w.WriteHeader(http.StatusInternalServerError)
+		_, _ = w.Write([]byte(err.Error()))
+	}
+}
```
x1m3, line 33: This code is repeated. Don't you think we could move the original one to an internal package and then import in both sides?

x1m3 (changes requested):

x1m3:
LGTM. Just added a request change to unify the middleware, but consider it more a suggestion than a request change. It if it not easy to move, let's go on and we iterate over it later.

internal/api_admin/server.go
martinsaporiti: Well, me neither too. I left those parameters for now until we will have more information regarding what we are going to need.

x1m3 (approved):

javip97 (approved):
LGTM

PR #191, opened 2023-03-01
User interface skeleton

x1m3:
UI skeleton.

- New docker in docker compose that runs code from /ui folder.

You can use the command make run-ui to start it.


infrastructure/local/docker-compose.yml
```diff
@@ -1,6 +1,15 @@
...
 services:
 
+  ui:
+    build:
+      context: ../../ui/
+      dockerfile: Dockerfile
+    ports:
+      - "8080:80"
+    environment:
+      - API=https://api-staging/polygonid/com/v1
```
adrifdez, line 11: Should be this `localhost:3001`?

martinsaporiti: Hi, it should be something like `http://backend:3002`. backend service is not merged yet, but after merging it we can call it with the service name (instead of the url)

x1m3: Yeah.. you are right.. Let's merge https://github.com/0xPolygonID/sh-id-platform/pull/190 before and i will change it

Makefile
```diff
@@ -62,6 +62,10 @@ run-arm:
 	COMPOSE_DOCKER_CLI_BUILD=1 KEY_STORE_TOKEN=$(TOKEN) DOCKER_FILE="Dockerfile-arm" $(DOCKER_COMPOSE_CMD) up -d platform
 	docker exec sh-id-platform-platform-1 ./migrate
 
+.PHONY: run-ui
+run-ui:
+	$(DOCKER_COMPOSE_CMD) up -d ui
```
elias-garcia, line 67: Are we assuming that the user has built the image before? Shouldn't we define a task to be able to build the UI as well?

ui/scripts/deploy.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+ENV_FILENAME="/app/.env"
```
elias-garcia, line 2: Wouldn't be easier to add a `WORDKIR /app` statement on top of the file instead of having the `/app` dir in every command below?

ui/README.md
```diff
@@ -0,0 +1,70 @@
...
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+
+## Available Scripts
+
+In the project directory, you can run:
+
+### `npm start`
+
+Runs the app in the development mode.\
+Open [http://localhost:3000](http://localhost:3000) to view it in your browser.
```
cneuro, line 12: Since the FE is built on [Vite](https://vitejs.dev), this will actually be http://localhost:5173/.

.github/CODEOWNERS
```diff
@@ -1,2 +1,5 @@
 # Global code owners
-*   @martinsaporiti @javip97 @x1m3
\ No newline at end of file
+*   @martinsaporiti @javip97 @x1m3
+
+# UI code owners
+/ui @adrifdez @cneuro @amonsosanz @elias-garcia
```
cneuro, line 5: Shouldn't this be managed via teams in [Settings](https://github.com/0xPolygonID/sh-id-platform/settings/access)?

Screenshot from 2023-03-01 17-07-39

ui/README.md
x1m3: This is the default readme for the react demo. All this code will change with the real UI code, including this file. I'm gonna remove it because nothing written  here is relevant.

BTW: Vite is still not installed. I left this part for FE team because you know much better than me how to use it.

.github/CODEOWNERS
x1m3: I agree, we could use groups. But there are people in the teams that maybe are not interested in being a code owner. I think raul jareño is in fe team. 
Let's discuss it or change it later to adapt to our needs.

infrastructure/local/docker-compose.yml
```diff
@@ -31,18 +40,18 @@ services:
...
+      #      - SH_ID_PLATFORM_ETHEREUM_RECEIPT_TIMEOUT=600s
+      #      - SH_ID_PLATFORM_ETHEREUM_MIN_GAS_PRICE=0
+      #      - SH_ID_PLATFORM_ETHEREUM_MAX_GAS_PRICE=1000000
+      #      - SH_ID_PLATFORM_ETHEREUM_RPC_RESPONSE_TIMEOUT=5s
+      #      - SH_ID_PLATFORM_ETHEREUM_WAIT_RECEIPT_CYCLE_TIME=30s
+      #      - SH_ID_PLATFORM_ETHEREUM_WAIT_BLOCK_CYCLE_TIME=30s
+      #      - SH_ID_PLATFORM_PROVER_SERVER_URL=http://localhost:8002
+      #      - SH_ID_PLATFORM_PROVER_TIMEOUT=600s
+      #      - SH_ID_PLATFORM_CIRCUIT_PATH=./pkg/credentials/circuits
       - SH_ID_PLATFORM_REDIS_URL=redis://@redis:6380/1
```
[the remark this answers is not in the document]
x1m3: It relies on env vars and config file. Env vars have precedence. We use here to help users to easily start the server with docker.

.github/CODEOWNERS
x1m3: Repo permissions have changed and I cannot see right now who is in which group

cneuro: Agreed, let's involve @mmilenkovic  in this to understand how to manage this going forward. First step is actually to remove Raul from the FE team, as this no longer makes sense.

ui/README.md
cneuro: Yeah, I am removing this entirely for now in favour of one README in root.

infrastructure/local/docker-compose.yml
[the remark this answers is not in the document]
cneuro: Discussion with Javier, to be resolved in next PR.

infrastructure/local/docker-compose.yml
x1m3: done

cneuro:
Perhaps this also needs to fix the `make run-ui` [error](https://0xpolygon.slack.com/archives/C036V06KE31/p1677774844745899?thread_ts=1677675808.439159&cid=C036V06KE31) and the updated Dockerfile for non-MacOS Linux, but these are already in the integration PR #197.

PR #196, opened 2023-03-02
chore: use golang:1.19 instead of ubuntu.

x1m3:
Use golang: 1.19 instead of ubuntu.

Image size reduced from 2.91GB to 2.19 GB


x1m3:
Closed because there the improvement is not impressive and this image could not be production ready.

PR #197, opened 2023-03-02
Added turnkey Issuer Node UI integration. Revised README.

cneuro:
Configuration requirements of UI env variables for:
- [x] Auth credentials (username & password) for UI-API endpoints
- [x] Auth credentials (username & password) for UI browser access (HTTP gate)
- [x] Issuer name
- [x] Issuer logo
- [x] Issuer DID

Main changes from Platform to `/ui`:
- Removed JWT handling & auth contexts
- Removed sign-in and its routes & components
- New (untyped) env var setup & handling


x1m3:
IMHO. This PR Must be broken into multiple ones. There are a lot of non related changes here that could have many implications.

cneuro:
> IMHO. This PR Must be broken into multiple ones. There are a lot of non related changes here that could have many implications.

Sure, let's discuss. What do you want to break off?

Ignoring the `config.toml.sample`, the only changes to existing config are using `docker compose` (instead of `docker-compose`) for compatibility, and using a functional `Dockerfile` for Linux (it was failing on non MacOS), provided by @martinsaporiti. Everything else is just UI, which we need to decide how to run.

x1m3:
I should only add UI. 

Changes in config.toml have implications for others. And it is not going to work as it is because the server is not adapted to this new format. And, there is another working task that is adding more changes to this config file.

Regarding docker compose versus docker-compose, it is ok. But it is a good thing to do in an isolated PR to track changes. 

Dockerfile is updated on the development branch. Maybe a better movement would be to merge develop on this branch.

cneuro:
> I should only add UI.

It does. But that requires at least several additions to the config, which are still pending.

> Changes in config.toml have implications for others. And it is not going to work as it is because the server is not adapted to this new format. And, there is another working task that is adding more changes to this config file.

I am happy to put the .sample file proposals somewhere else - it's just meant for discussion for now. We don't need to commit it, let's find a way to organize this in our meeting.

> Regarding docker compose versus docker-compose, it is ok. But it is a good thing to do in an isolated PR to track changes.

This literally doesn't work with the latest Docker on my Linux machine though, and it doesn't change functionality AFAIK.

> Dockerfile is updated on the development branch. Maybe a better movement would be to merge develop on this branch.

It already is :) The reason we're seeing it here is because this PR is pointing to `ID-265-welcome-frontend`.

README.md
```diff
@@ -1,84 +1,119 @@
...
+
+#### Setup for Docker-only
+
+1. Copy `.env-api.sample` as `.env-api` and `.env-issuer.sample` as `.env-issuer`. Please see the [configuration](#configuration) section for more details.
+2. Run `make up`. This launches 3 containers with Postgres, Redis and Vault. Ignore the warnings about variables, since those are set up in the next step.
+3. **If you are on an Apple Silicon chip (e.g. M1/M2), run `make run-arm`**. Otherwise, run `make run`. This starts Docker containers for the issuer application.
+4. Follow the [steps](#adding-ethereum-private-key-to-the-vault) for adding an Ethereum private key to the Vault.
+5. Open <http://localhost:3001> in a browser (or whatever was set in the `[Server] URL` config entry). This shows an admin interface for documentation and credentials issuer setup.
+6. _(Optional)_ To run the UI with its own API, first copy `.env-ui.sample` as `.env-ui`. Please see the [configuration](#configuration) section for more details.
+7. _(Optional)_ Run `make-run-ui` (or `make-run-ui-arm` on Apple Silicon) to have the Web UI available on <http://localhost:80>. Its HTTP auth credentials are set in `.env-ui`.
```
amonsosanz, line 32: these commands are not correct:

`make run-ui` and `make run-arm-ui`

cneuro: Oops, thanks!

amonsosanz: The port is also incorrect: `http://localhost:8080/`

cneuro: > The port is also incorrect: http://localhost:8080/

This is fixed.

> I'd consider changing the command make run-arm-ui to make run-ui-arm

It's already `make run-ui-arm`.

x1m3 (approved):

martinsaporiti (approved):

PR #198, opened 2023-03-02
Id 222 ship add optional logo url issuer name in the config toml

martinsaporiti:


internal/config/config.go
```diff
@@ -114,6 +114,15 @@ type HTTPBasicAuth struct {
 	Password string `mapstructure:"Password" tip:"Basic auth password"`
 }
 
+// Admin - Admin backend service configuration.
+type Admin struct {
+	ServerPort    int           `mapstructure:"ServerPort"`
+	HTTPAdminAuth HTTPAdminAuth `mapstructure:"HTTPAdminAuth"`
+	IssuerName    string        `mapstructure:"IssuerName"`
+	IssuerLogo    string        `mapstructure:"IssuerLogo"`
+	IssuerDID     string        `mapstructure:"IssuerDID"`
```
x1m3, line 123: We should add the "tip" tag. This is used by the experimental configurator to give a tip to the user about the purpose of this field

x1m3 (changes requested):

internal/config/config.go
martinsaporiti: done

javip97 (approved):

x1m3 (approved):

PR #202, opened 2023-03-09
feat: auth qr code and authentication callback

javip97:
This RP covers the following aspects:
* Authentication qr generation
* Callback of authentication
* Connection creation


martinsaporiti (approved):
LGTM. Nice Work!

.golangci.yml
```diff
@@ -71,7 +71,7 @@ linters-settings:
       - name: argument-limit
         severity: warning
         disabled: false
-        arguments: [ 9 ]
+        arguments: [ 12 ]
```
x1m3, line 74: OMG ;-)

internal/db/schema/migrations/20230308093840_add_connections_table.sql
```diff
@@ -0,0 +1,22 @@
...
+    user_doc    jsonb NULL,
+    created_at             TIMESTAMP WITH TIME ZONE DEFAULT current_timestamp,
+    modified_at             TIMESTAMP WITH TIME ZONE default current_timestamp,
+    CONSTRAINT connections_issuer_user_key UNIQUE (issuer_id, user_id)
+);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE connections DROP CONSTRAINT connections_issuer_user_key;
```
x1m3, line 20: We should recreate the original connections table here

internal/repositories/connections.go
```diff
@@ -0,0 +1,27 @@
...
+func NewConnections() ports.ConnectionsRepository {
+	return &connections{}
+}
+
+// Save stores in the database the given connection and updates the modified at in case already exists
+func (c *connections) Save(ctx context.Context, conn db.Querier, connection *domain.Connection) error {
+	sql := `INSERT INTO connections (issuer_id, user_id, issuer_doc, user_doc)
+			VALUES($1, $2, $3, $4) ON CONFLICT (issuer_id, user_id) DO
+			UPDATE SET issuer_id=$1, user_id=$2, issuer_doc=$3, user_doc=$4,
+			           modified_at = now();`
```
x1m3, line 23: I think it is better to use connection.CreatedAt instead of now().

internal/db/schema/migrations/20230308093840_add_connections_table.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE connections DROP CONSTRAINT connections_managed_identifier_third_party_identifier_key;
+DROP TABLE IF EXISTS connections;
+CREATE TABLE connections
+(
+    id                     uuid                     DEFAULT gen_random_uuid(),
```
x1m3, line 7: Do not generate ids at repo level.

x1m3 (changes requested):

.golangci.yml
javip97: we can split that identity service into two in order to lower down that number

internal/db/schema/migrations/20230308093840_add_connections_table.sql
javip97: right, nice point

internal/repositories/connections.go
javip97: agree, thanks

x1m3 (approved):

PR #215, opened 2023-03-14
chore: check did when ui api starts

martinsaporiti:


internal/core/ports/identity_service.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // IdentityService is the interface implemented by the identity service
 type IdentityService interface {
+	GetByDID(ctx context.Context, identifier *core.DID) (*domain.Identity, error)
```
x1m3, line 16:
```suggestion
	GetByDID(ctx context.Context, identifier core.DID) (*domain.Identity, error)
```
IMHO, it is better to pass a core.DID and not a DID, as the parameter is mandatory and there is no performance improvement in passing a pointer to an array of bytes.

This pointer could also be removed from other functions calling this method.

x1m3 (changes requested):
Good!. Just 2 small request changes.

x1m3 (approved):

PR #218, opened 2023-03-14
Integrated latest changes to Platform for final migration to issuer node UI. Updated & improved README.

cneuro:


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
amonsosanz, line 1: What is this required for?

cneuro: Indeed. Fixed.

ui/src/utils/adapters.ts
```diff
@@ -156,3 +158,61 @@ export const stringBoolean = StrictSchema<string, boolean>()(
     }
   })
 );
+
+// TODO - are the following constants and functions adapters? Parsers? Should they live in "utils/json.ts"? They look out of place here.
```
amonsosanz, line 162: I think they are ok here. This is a generic adapter-utils module. I'd not create another adapter-utils module just to hold `jsonLiteralParser`, `jsonParser`, and `getJsonFromUrl` but this is, of course, subjective.

cneuro: What are the criteria of choosing where to add these types of functions and types? How are some an "adapter", but not a "parser"? In `adapters/api/parsers`, we have `jsonLdTypeParser`, but here we have `jsonLiteralParser`. We also have `iden3JsonLdTypeParser`, but we also have a `utils/iden3.ts`.

I'm confused by the entire file structure and the amount of similarly named functions, types and constants in all these different places and I can't discern what goes where from a logical pov.

cneuro: Thanks for the call and discussion. All further action will be taken as part of [this issue](https://polygonid.atlassian.net/browse/PID-395?atlOrigin=eyJpIjoiNTI1ZGE3OTIxMmZjNDI0ZjgxNDRiOGMxOWU4MmQ5MTciLCJwIjoiaiJ9), as discussed.

ui/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
```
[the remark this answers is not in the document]
cneuro: Fixed and consolidated.

ui/src/components/shared/SiderMenu.tsx
```diff
@@ -35,12 +40,19 @@ export function SiderMenu() {
           items={[
             {
               icon: <IconSchema />,
+              // TODO - these keys need to be typed.
```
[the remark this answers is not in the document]
amonsosanz: I assume the problem you want to solve is a key collision, right? If that is the case, I'm not getting how a union type would help there, since the type would not save us from:

 * Using any other arbitrary string as key
 * Using repeated values that honor the type
 
Or are you maybe solving another problem that I have not thought of?

ui/src/utils/adapters.ts
amonsosanz: Thanks to you! As we commented in the meet the name "adapters" was proposed by @elias-garcia and I think it comes from the [hexagonal architecture](https://en.wikipedia.org/wiki/Hexagonal_architecture_(software)).

To be honest, I never liked it much. I was used to naming the layer that integrates external parties into the core of the app (parsing/serializing data) "persistence", but in the end, it's just a name and has no relevance at all.

ui/src/components/shared/SiderMenu.tsx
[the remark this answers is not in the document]
cneuro: Thanks for helping come up with the optimal solution!

amonsosanz (approved):

PR #222, opened 2023-03-15
feat: Get All Schemas endpoint

x1m3:
Basic implementation of Get All Schemas endpoint. It returns all schemas, without filtering.


javip97 (approved):
LGTM

PR #225, opened 2023-03-16
Added persistent dismissal to Explainers.

cneuro:


ui/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,15 @@
+import ls from "localstorage-slim";
```
amonsosanz, line 1: Do we need a lib for this? Why don't we get the value directly from the local storage?

ui/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,15 @@
...
+  key: string,
+  defaultValue: T
+): [T, React.Dispatch<React.SetStateAction<T>>] {
+  const [value, setValue] = useState(ls.get<T>(key) ?? defaultValue);
+
+  useEffect(() => {
+    ls.set(key, value);
+  }, [key, value]);
+
+  return [value, setValue];
```
amonsosanz, line 14: Assuming we agree to use the helpers from the bridge, does this hook still make sense?

ui/src/components/shared/Explainer.tsx
```diff
@@ -1,21 +1,28 @@
...
   image: string;
   title: string;
 }) {
+  const [isShowing, setShowing] = useLocalStorage(`explainer-${kebabCase(title)}`, true);
+
   const { xl } = Grid.useBreakpoint();
   const { label, url } = CTA;
 
+  if (!isShowing) {
+    return null;
```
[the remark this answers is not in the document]
cneuro: Why? The whole point is that every Explainer will only show if it is not dismissed. This logic is now handled via an Explainer's own special flag in localStorage. The parent doesn't care nor should it need to - it saves lifted state management. If we would want an explainer without a dismiss functionality, we an always make that a prop ike `canDismiss={false}`.

ui/src/components/shared/Explainer.tsx
```diff
@@ -1,21 +1,28 @@
...
   image,
   title,
 }: {
   CTA: { label: string; url: string };
   description: string;
-  handleDismiss: () => void;
   image: string;
   title: string;
 }) {
+  const [isShowing, setShowing] = useLocalStorage(`explainer-${kebabCase(title)}`, true);
```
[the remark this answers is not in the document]
cneuro: I suppose I am not expecting the user to intentionally break their own application by mangling its localStorage. But it's true, better to throw an error rather than having the app crash.

ui/src/hooks/useLocalStorage.ts
cneuro: This lib was added to avoid having to deal with `JSON.parse` which is a typing hellscape. With this, we can just set `ls.get<T>(VALUE)` to avoid defining all our own parsers and typeguards.

But it looks like this already achieved by get & set `StorageByKey`, so I will look into that.

ui/src/hooks/useLocalStorage.ts
cneuro: I would combine the two, because the point of the hook is that it sets the state as well without having to additionally handle it in the component.

ui/src/components/shared/Explainer.tsx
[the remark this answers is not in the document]
cneuro: Amended as discussed.

ui/src/components/shared/Explainer.tsx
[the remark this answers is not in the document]
cneuro: Amended as discussed.

ui/src/hooks/useLocalStorage.ts
cneuro: Removed as discussed.

ui/src/hooks/useLocalStorage.ts
cneuro: Removed as discussed.

ui/src/components/schemas/Schemas.tsx
```diff
@@ -43,15 +39,14 @@ export function Schemas() {
...
-            image="/images/illustration-explainer.svg"
-            title="Credential schemas explained"
-          />
-        )}
+        <Explainer
+          CTA={{ label: "Learn more", url: TUTORIALS_URL }}
+          description="Learn about schema types, attributes, naming conventions, data types and more."
+          image="/images/illustration-explainer.svg"
+          isDismissable
+          localStorageKey="explainerSchemas"
```
amonsosanz, line 47: I'd consider removing `isDismissable` and typing `localStorageKey` as `string | undefined`.
Like this, we could infer that whenever a `localStorageKey` is present, the explainer is dismissable and it's not otherwise.

The current API forces us to provide a `localStorageKey` even when `isDismissable` is set as false, which makes no sense.

cneuro: Fair point. Fixed.

ui/src/components/shared/Explainer.tsx
```diff
@@ -30,9 +46,17 @@ export function Explainer({
...
           )}
 
-          <Button onClick={handleDismiss} type="primary">
-            Dismiss
-          </Button>
+          {isDismissable && (
+            <Button
+              onClick={() => {
+                setShowing(false);
+                setStorageByKey({ key: localStorageKey, value: false });
```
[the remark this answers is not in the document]
cneuro: Great, thanks!

PR #233, opened 2023-03-17
fix: returning valid error message in invalid requests

javip97:
This PR returns a valid error response message for request validations

before:
`Invalid format for parameter id: error unmarshalling '8edd8112-c415-11ed-b036-debe37e1cbd612345asd\n' text as *uuid.UUID: invalid urn prefix: \"8edd8112-\"`

after:
```
{
    "message": "Invalid format for parameter id: error unmarshalling '8edd8112-c415-11ed-b036-debe37e1cbd612345asd\n' text as *uuid.UUID: invalid urn prefix: \"8edd8112-\""
}
```


martinsaporiti (approved):
very good!!

x1m3 (approved):

PR #241, opened 2023-03-20
Parse UI env variables with zod

amonsosanz:
Resolves: https://polygonid.atlassian.net/browse/PID-389


ui/src/styles/index.scss
```diff
@@ -200,6 +200,10 @@
   width: 16px;
 }
 
+.ant-result-subtitle {
```
cneuro, line 203: This is part of your [other](https://github.com/0xPolygonID/sh-id-platform/pull/236) PR?

ui/src/adapters/env.ts
```diff
@@ -0,0 +1,46 @@
+import { z } from "zod";
+
+import * as domain from "src/domain";
```
cneuro, line 3: We should avoid wild card imports and call either the interface here or the one in `domain` something else, or reassign it locally.

ui/src/adapters/env.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface Env {
+  VITE_API_PASSWORD: string;
+  VITE_API_URL: string;
+  VITE_API_USERNAME: string;
+  VITE_ISSUER_DID: string;
+  VITE_ISSUER_LOGO?: string;
+  VITE_ISSUER_NAME?: string;
+}
+
+export const envParser = StrictSchema<Env, domain.Env>()(
```
cneuro, line 15: We can't just run this parser on startup and have them available as global vars? Not sure why they need to be handled in a React context if they're static, it only seems to add complexity and prop/param drilling.

ui/src/contexts/env.context.tsx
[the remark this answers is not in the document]
amonsosanz: It's the context for the env. We might need more contexts in the future that would be in separate modules. I'm following the same pattern we have in the [zkEVM Bridge UI](https://github.com/0xPolygonHermez/zkevm-bridge-ui/tree/develop/src/contexts).

ui/src/styles/index.scss
amonsosanz: Yep, I've cherry-picked the same commit.

ui/src/adapters/env.ts
amonsosanz: > We should avoid wild card imports and call the different types of `Env` that are declared differently.

Why? The namespace is precisely there to provide the context.

> They're exported and should be addressable uniquely

They are addressable uniquely when properly namespaced.

> to avoid confusion and inconsistencies

Can you extend on what confusion and inconsistencies you see with the namespaces?

ui/src/contexts/env.context.tsx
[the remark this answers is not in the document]
cneuro: I'm sorry, I meant to say, why not name it just `env.tsx`? We don't have any sort of file naming convention using `.` in `src`.

ui/src/adapters/env.ts
cneuro: A developer can write `import * as foo from "src/domain";`. It goes against the spirit of the linter enforcing named exports, since it leaves ambiguity.

Further, it makes it difficult to know which `Env` I am looking for and reliant on the IDE for hints. Lastly, we don't do this anywhere else, everything is a uniquely named export.

ui/src/styles/index.scss
cneuro: Why though? What do you need it for here?

ui/src/contexts/env.context.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const value = useMemo(() => {
+    return env;
+  }, [env]);
+
+  return value?.success ? (
+    <envContext.Provider value={value.data} {...props} />
+  ) : !value ? null : (
+    <ErrorResult
+      error={[
+        "An error occurred while trying to parse then env:\n",
```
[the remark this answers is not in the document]
amonsosanz: Ok, done.

ui/src/contexts/env.context.tsx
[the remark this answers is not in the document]
amonsosanz: Ah sure! Done.

ui/src/styles/index.scss
amonsosanz: To properly render the multiline env parsing errors:

image

ui/src/adapters/env.ts
amonsosanz: > A developer can write `import * as foo from "src/domain";`.

I don't see any problem with this, it's up to the developer to decide the proper namespace in the context of the importing module, but ok will change it. It's not a big deal.

ui/src/adapters/env.ts
cneuro: I didn't realize that the app was rendering an error, I thought it would just fail. This makes sense then.

ui/src/styles/index.scss
cneuro: My question is: why is this duplicated in this PR?

ui/src/adapters/env.ts
amonsosanz: Changed

cneuro: Hmm, let's change it to only import what we want. I think we should remain consistent and strict with naming.

ui/src/contexts/env.context.tsx
```diff
@@ -0,0 +1,58 @@
...
+  PropsWithChildren,
+  createContext,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
+
+import { z } from "zod";
+import * as adapter from "src/adapters/env";
```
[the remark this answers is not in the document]
amonsosanz: Changed

ui/src/adapters/env.ts
cneuro: Can we also change it in `schemas.ts`? Just to remain consistent and also strict in naming, importing only what we use.

ui/src/styles/index.scss
amonsosanz: Because otherwise in this PR we would see this, which is incorrect:

image

ui/src/adapters/env.ts
amonsosanz: Not in this PR. This is a massive change that would imply a lot of interface renames, plus I don't agree with the change and would like to discuss it first in the context of a dedicated issue. Does it sound fair?

cneuro (approved):

ui/src/adapters/api/credentials.ts
```diff
@@ -2,22 +2,20 @@ import axios from "axios";
...
-  API_PASSWORD,
-  API_URL,
-  API_USERNAME,
-  ISSUER_DID,
-  QUERY_SEARCH_PARAM,
-} from "src/utils/constants";
+import { QUERY_SEARCH_PARAM } from "src/utils/constants";
 import { StrictSchema } from "src/utils/types";
 
+const buildAuthorizationHeader = (env: Env) =>
```
cneuro, line 16: Since we'll need this for all endpoints, it should probably live in `utils`.

ui/src/adapters/env.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const envParser = StrictSchema<EnvInput, Env>()(
+  z
+    .object({
+      VITE_API_PASSWORD: z.string().min(1),
+      VITE_API_URL: z.string().url(),
+      VITE_API_USERNAME: z.string().min(1),
+      VITE_ISSUER_DID: z.string(),
+      VITE_ISSUER_LOGO: z.string().optional(),
+      VITE_ISSUER_NAME: z.string().optional(),
```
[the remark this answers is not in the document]
amonsosanz: ah! ok. Changed.

ui/src/adapters/api/credentials.ts
amonsosanz: Yep, I also thought that but, in what utils module? I have resisted the idea of creating another module in utilities just for this function. Should I?

cneuro: It's a good question. Let's leave it for now, I will get back to this in my PR where we need a cohesive approach to env provisioning.

PR #250, opened 2023-03-22
Proposal to change Context to HoC

adrifdez:
The idea is to remove the necessity of using hooks with the statics values of the environment


ui/src/components/shared/EnvHoC.tsx
```diff
@@ -0,0 +1,39 @@
+import { FC, PropsWithChildren } from "react";
+
+import { envParser } from "src/adapters/env";
+import { ErrorResult } from "src/components/shared/ErrorResult";
+import { Env } from "src/domain";
+import { processZodError } from "src/utils/adapters";
+
+const envParsed = envParser.safeParse(import.meta.env);
```
amonsosanz, line 8: I'd rename it to `parsedEnv`

ui/src/components/shared/EnvHoC.tsx
```diff
@@ -0,0 +1,39 @@
...
+      },
+      issuer: {
+        did: "",
+        name: "",
+      },
+    };
+
+export const authorization = `Basic ${env.api.username}:${env.api.password}`;
+
+const EnvHoC: FC<PropsWithChildren> = ({ children }) => {
```
cneuro, line 25: Why is this a constant and not a named function like all other components?

ui/src/components/shared/EnvHoC.tsx
```diff
@@ -0,0 +1,39 @@
...
+        url: "",
+        username: "",
+      },
+      issuer: {
+        did: "",
+        name: "",
+      },
+    };
+
+export const authorization = `Basic ${env.api.username}:${env.api.password}`;
```
cneuro, line 23: Since this is exported, it would be clearer to call this something like `apiAuthHeader`.

amonsosanz (approved):

PR #253, opened 2023-03-22
feat: adding revoke all claims of a connection

javip97:
Revoking all claims of a connection will revoke all the non revoked claims of the given connection.

The revoking claims works as a transaction, or every claim get revoked or non get revoked

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -136,6 +136,7 @@ func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObje
 func (s *Server) RevokeClaim(ctx context.Context, request RevokeClaimRequestObject) (RevokeClaimResponseObject, error) {
 	did, err := core.ParseDID(request.Identifier)
 	if err != nil {
+		log.Warn(ctx, "revoke claim invalid did", "err", err, "req", request)
 		return RevokeClaim400JSONResponse{N400JSONResponse{err.Error()}}, nil
 	}
 
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -134,7 +134,12 @@ func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObje
 
 // RevokeClaim is the revocation claim controller
 func (s *Server) RevokeClaim(ctx context.Context, request RevokeClaimRequestObject) (RevokeClaimResponseObject, error) {
-	if err := s.claimService.Revoke(ctx, request.Identifier, uint64(request.Nonce), ""); err != nil {
+	did, err := core.ParseDID(request.Identifier)
+	if err != nil {
+		return RevokeClaim400JSONResponse{N400JSONResponse{err.Error()}}, nil
+	}
+
+	if err := s.claimService.Revoke(ctx, *did, uint64(request.Nonce), ""); err != nil {
 		if errors.Is(err, repositories.ErrClaimDoesNotExist) {
 			return RevokeClaim404JSONResponse{N404JSONResponse{
 				Message: "the claim does not exist",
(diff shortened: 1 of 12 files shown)


internal/repositories/claims.go
```diff
@@ -403,6 +403,44 @@ func (c *claims) GetAllByIssuerID(ctx context.Context, conn db.Querier, identifi
...
+			`
+
+	rows, err := conn.Query(ctx, query, connID.String(), issuerID.String())
+	if err != nil {
+		if err == pgx.ErrNoRows {
+			return []*domain.Claim{}, nil // connections can have 0 credentials
+		}
+
+		return nil, err
+	}
```
x1m3, line 438:
```suggestion
	if err != nil && err != pgx.ErrNoRows {
		return nil, err
	}
```

internal/api_admin/server.go
```diff
@@ -325,3 +300,38 @@ func (s *Server) PublishState(ctx context.Context, request PublishStateRequestOb
 		TxID:               publishedState.TxID,
 	}, nil
 }
+
+// RevokeConnectionCredentials revoke all the non revoked credentials of the given connection
+func (s *Server) RevokeConnectionCredentials(ctx context.Context, request RevokeConnectionCredentialsRequestObject) (RevokeConnectionCredentialsResponseObject, error) {
+	err := s.claimService.RevokeAllFromConnection(ctx, request.Id, s.cfg.APIUI.IssuerDID)
+	if err != nil {
```
x1m3, line 307: log the error

internal/api_admin/server.go
```diff
@@ -296,7 +271,7 @@ func (s *Server) CreateCredential(ctx context.Context, request CreateCredentialR
 
 // RevokeCredential - revokes a credential per a given nonce
 func (s *Server) RevokeCredential(ctx context.Context, request RevokeCredentialRequestObject) (RevokeCredentialResponseObject, error) {
-	if err := s.claimService.Revoke(ctx, s.cfg.APIUI.IssuerDID.String(), uint64(request.Nonce), ""); err != nil {
+	if err := s.claimService.Revoke(ctx, s.cfg.APIUI.IssuerDID, uint64(request.Nonce), ""); err != nil {
```
x1m3, line 274: log the errror

internal/api/server.go
```diff
@@ -134,7 +134,12 @@ func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObje
 
 // RevokeClaim is the revocation claim controller
 func (s *Server) RevokeClaim(ctx context.Context, request RevokeClaimRequestObject) (RevokeClaimResponseObject, error) {
-	if err := s.claimService.Revoke(ctx, request.Identifier, uint64(request.Nonce), ""); err != nil {
+	did, err := core.ParseDID(request.Identifier)
+	if err != nil {
+		return RevokeClaim400JSONResponse{N400JSONResponse{err.Error()}}, nil
```
x1m3, line 139: log the error

x1m3 (changes requested):
LGTM.  Just some small comments regarding logs.

internal/repositories/claims.go
javip97: agree, I will change the defer as well 

```
defer rows.Close()
	
if err != nil && err != pgx.ErrNoRows {
	return nil, err
}
```

internal/api_admin/server.go
javip97: thanks

internal/api/server.go
javip97: thanks

internal/api_admin/server.go
javip97: thanks

PR #254, opened 2023-03-22
feat: Get Credentials endpoint

x1m3:


internal/api_admin/server.go
```diff
@@ -225,6 +227,37 @@ func (s *Server) GetCredential(ctx context.Context, request GetCredentialRequest
...
+// GetCredentials returns a collection of credentials that matches the request.
+func (s *Server) GetCredentials(ctx context.Context, request GetCredentialsRequestObject) (GetCredentialsResponseObject, error) {
+	filter := &ports.ClaimsFilter{}
+	if request.Params.Type != nil {
+		switch GetCredentialsParamsType(strings.ToLower(string(*request.Params.Type))) {
+		case Revoked:
+			filter.Revoked = common.ToPointer(true)
+		case Expired:
+			filter.ExpiredOn = common.ToPointer(time.Now())
+		}
```
javip97, line 239: I may have added here a default case that throws an error that propagates into a 400 bad request so the issuer is aware that the provided filter is not valid, giving him back the valid filters, in this case "revoked" and "expired"

javip97 (approved):
LGTM, nice Job

PR #256, opened 2023-03-23
Improve organization of adapters parsers types and their files

amonsosanz:
Implements: https://polygonid.atlassian.net/browse/PID-395


ui/src/styles/theme.ts
```diff
@@ -5,7 +5,7 @@ import { StyleVariables } from "src/domain";
 import variables from "src/styles/variables.module.scss";
 import { StrictSchema } from "src/utils/types";
 
-export const parseStyleVariables = StrictSchema<StyleVariables>()(
+const parseStyleVariables = StrictSchema<StyleVariables>()(
```
cneuro, line 8: If we only ever use this once, should it just be in `parsedStyleVariables`? Ditto for other instances like this.

ui/src/adapters/api/index.ts
```diff
@@ -0,0 +1,84 @@
...
+);
+
+export function buildAPIError(error: unknown): APIError {
+  if (axios.isCancel(error)) {
+    return { message: error.toString(), status: HTTPStatusError.Aborted };
+  }
+
+  if (axios.isAxiosError(error)) {
+    try {
+      // This is a Polygon ID API error.
```
cneuro, line 65: This should say "UI API error" instead.

ui/src/styles/theme.ts
amonsosanz: I'm not sure I understand the comment. If you refer to moving the type `StyleVariables` into this module I can't agree more. Done. If you meant something else, can you please reformulate the comment?

ui/src/utils/error.ts
```diff
@@ -0,0 +1,26 @@
+import z from "zod";
+
+export function processZodError<T>(error: z.ZodError<T>, init: string[] = []): string[] {
```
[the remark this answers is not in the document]
amonsosanz: I think the current names are correct:

 * `buildAPIError` is indeed building an object of type `APIError`. I think the prefix `process` would not apply to it.
 * `processZodError` is not building any type, it's just "processing" a zod error to extract a list of human-readable strings from it. I think the prefix `build` would not apply to it.

ui/src/adapters/api/index.ts
amonsosanz: Changed

ui/src/styles/theme.ts
amonsosanz: Ahh! oc we can parse the value directly after defining the parser.

Also, we don't need to use `safeParse` to throw a string in the `if` block below. It's better to use `parse` and let `zod` throw with a more meaningful error. A different discussion is whether we should throw like this, but I was not involved in this implementation and can't tell for sure why all this was architectured this way.

Changed.

cneuro (approved):

adrifdez (approved):

PR #257, opened 2023-03-23
feat: adding the option to revoke & delete credentials in del conn endpoint

javip97:
This PRs covers the possibility of revoking and deleting the credentials of a given connection.


martinsaporiti (approved):
LGTM

PR #258, opened 2023-03-23
Pid 180 create credential link issue credential link

martinsaporiti:


internal/api_admin/server.go
```diff
@@ -323,6 +328,81 @@ func (s *Server) RevokeConnectionCredentials(ctx context.Context, request Revoke
...
+		if err != nil {
+			return CreateLink400JSONResponse{N400JSONResponse{Message: "wrong expirationDate attribute format"}}, nil
+		}
+		if isBeforeTomorrow(t) {
+			return CreateLink400JSONResponse{N400JSONResponse{Message: "expirationDate cannot be a date before tomorrow"}}, nil
+		}
+	}
+
+	if request.Body.ClaimLinkExpiration != nil {
+		if !isClaimLinkValid(*request.Body.ClaimLinkExpiration, expiresAt) {
```
javip97, line 347: This should be changed as well

internal/api_admin/server.go
```diff
@@ -323,6 +328,81 @@ func (s *Server) RevokeConnectionCredentials(ctx context.Context, request Revoke
...
+		if *request.Body.LimitedClaims <= 0 {
+			return CreateLink400JSONResponse{N400JSONResponse{Message: "limitedClaims must be higher than 0"}}, nil
+		}
+		lc = common.ToPointer(int32(*request.Body.LimitedClaims))
+	}
+
+	// Todo improve validations errors
+	link := domain.NewLink(s.cfg.APIUI.IssuerDID, lc, request.Body.ClaimLinkExpiration, request.Body.SchemaID, request.Body.SignatureProof, request.Body.MtProof, attrs)
+	id, err := s.linkService.Save(ctx, *link)
+	if err != nil {
```
javip97, line 375: add some logs here

internal/api_admin/server.go
```diff
@@ -323,6 +328,81 @@ func (s *Server) RevokeConnectionCredentials(ctx context.Context, request Revoke
 	return RevokeConnectionCredentials202JSONResponse{Message: "Credentials revocation request sent"}, nil
 }
 
+// CreateLink - creates a link for issuing a credential
+func (s *Server) CreateLink(ctx context.Context, request CreateLinkRequestObject) (CreateLinkResponseObject, error) {
+	var expiresAt *time.Time
+	if request.Body.ExpirationDate != nil {
+		layout := "2006/01/02"
+		t, err := time.Parse(layout, *request.Body.ExpirationDate)
```
x1m3, line 336: It is better to use date format modifier in the yaml so we don't need to parse it here and use standard dates.

internal/db/schema/migrations/20230322114550_add_links_table.sql
```diff
@@ -0,0 +1,24 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE links
+(
+    id                              UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
x1m3, line 5:
```suggestion
    id                              UUID PRIMARY NOT NULL,
```

It is better to generate ids in the domain, instead of relying on database generation. This is what we've been doing in the last tables we've created.

internal/core/domain/link.go
```diff
@@ -0,0 +1,68 @@
...
+	SchemaID                 uuid.UUID
+	CredentialExpiration     *time.Time
+	CredentialSignatureProof bool
+	CredentialMTPProof       bool
+	CredentialAttributes     []CredentialAttributes
+	Active                   bool
+}
+
+// NewLink - Constructor
+func NewLink(IssuerDID core.DID, maxIssuance *int32, validUntil *time.Time, schemaID uuid.UUID, CredentialSignatureProof bool, CredentialMTPProof bool, credentialAttributes []CredentialAttributes) *Link {
```
x1m3, line 38:
```suggestion
func NewLink(IssuerDID core.DID, maxIssuance *int, validUntil *time.Time, schemaID uuid.UUID, CredentialSignatureProof bool, CredentialMTPProof bool, credentialAttributes []CredentialAttributes) *Link {
```
I think it is better to use an int instread of an int32. Less code conversions and probably bigger range.

internal/core/services/link.go
```diff
@@ -0,0 +1,26 @@
...
+	linkRepository ports.LinkRepository
+}
+
+// NewLinkService - constructor
+func NewLinkService(linkRepository ports.LinkRepository) ports.LinkService {
+	return &Link{linkRepository: linkRepository}
+}
+
+// Save - save a new credential
+func (ls *Link) Save(ctx context.Context, link domain.Link) (*uuid.UUID, error) {
```
x1m3, line 23: I think it is better to pass parameters or a DTO request and create the domain.Link entity inside the Save method.  Also, in can be more useful to return the domain.Link entity created and not only the id. So the caller can use the whole entity if needed.

x1m3 (changes requested):
I added some request changes. We can discuss it between. We need to agree in how to generate id. I prefer to do it inside the domain and not let the DB to generate for us.

internal/api_admin/server.go
martinsaporiti: Can `ClaimLinkExpiration` be set before today?

internal/db/schema/migrations/20230322114550_add_links_table.sql
martinsaporiti: > I added some request changes. We can discuss it between the team. We need to agree in how to generate ids. I prefer to do it inside the domain and not let the DB to generate for us.

I dont agree with that. But I will change it. I don't see a good reason.

internal/core/services/link.go
martinsaporiti: in this case I dont see the why is better the dto or the parameters. If we pass a dto, the service must to create the domain link (only that). Same if we pass the parameters. DTOS are useful in other contexts like credential creation. In this context I dont see the benefit. anyway i will change it. Regarding the response, its ok

internal/core/domain/link.go
martinsaporiti: ok

internal/core/domain/link.go
```diff
@@ -0,0 +1,70 @@
...
+	SchemaID                 uuid.UUID
+	CredentialExpiration     *time.Time
+	CredentialSignatureProof bool
+	CredentialMTPProof       bool
+	CredentialAttributes     []CredentialAttributes
+	Active                   bool
+}
+
+// NewLink - Constructor
+func NewLink(id uuid.UUID, issuerDID core.DID, maxIssuance *int, validUntil *time.Time, schemaID uuid.UUID, credentialExpiration *time.Time, CredentialSignatureProof bool, CredentialMTPProof bool, credentialAttributes []CredentialAttributes) *Link {
```
x1m3, line 38: IMHO, The ID should get be initiated here, in the constructor.  Not as a parameter. This is the point where the Link entity is created, so this is a good point to give an ID.

internal/core/ports/link_service.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/google/uuid"
+	core "github.com/iden3/go-iden3-core"
+
+	"github.com/polygonid/sh-id-platform/internal/core/domain"
+)
+
+// LinkService - the interface that defines the available methods
+type LinkService interface {
+	Save(ctx context.Context, ID uuid.UUID, did core.DID, maxIssuance *int, validUntil *time.Time, schemaID uuid.UUID,
```
x1m3, line 15: Related to a previous comment regarding ID. The ID should not be a param here. ID should be generated when constructing the new Link entity inside the Save function.  The best point is in the core.Domain constructor or very close to it.

x1m3 (changes requested):

internal/db/schema/migrations/20230322114550_add_links_table.sql
x1m3: Generating IDs in the domain isolates you from repository implementation, You can mock the repo or reimplement it in another persistence layer like Redis, Mongo, etc.. and the  id generation will not change if done in the domain.

javip97 (approved):
LGTM

x1m3 (approved):

PR #260, opened 2023-03-24
Fix makefile restart-ui

adrifdez:
This solves the problem of running the command make restart-ui


martinsaporiti (approved):
Nice, thanks!

PR #261, opened 2023-03-24
fix: env variable ResolverPrefix not binded

javip97:


internal/config/config.go
```diff
@@ -309,6 +309,7 @@ func bindEnv() {
 	_ = viper.BindEnv("Ethereum.RPCResponseTimeout", "ISSUER_ETHEREUM_RPC_RESPONSE_TIMEOUT")
 	_ = viper.BindEnv("Ethereum.WaitReceiptCycleTime", "ISSUER_ETHEREUM_WAIT_RECEIPT_CYCLE_TIME")
 	_ = viper.BindEnv("Ethereum.WaitBlockCycleTime", "ISSUER_ETHEREUM_WAIT_BLOCK_CYCLE_TIME")
+	_ = viper.BindEnv("Ethereum.ResolverPrefix", "ISSUER_ETHEREUM_RESOLVER_PREFIX")
```
adrifdez, line 312: You should add it to the sample file

javip97: right, thanks!

x1m3 (approved):

PR #264, opened 2023-03-27
Avoid unknown error and simplify state

amonsosanz:


ui/src/components/schemas/SchemaDetails.tsx
```diff
@@ -155,14 +147,39 @@ export function SchemaDetails() {
...
-        if (
-          isAsyncTaskDataAvailable(apiSchema) &&
-          isAsyncTaskDataAvailable(schema) &&
-          isAsyncTaskDataAvailable(jsonLdType) &&
-          rawJsonLdContext &&
-          rawJsonSchema
-        ) {
+        if (hasAsyncTaskFailed(apiSchema)) {
+          return (
+            <Card style={{ margin: "auto", maxWidth: CARD_WIDTH }}>
```
cneuro, line 152: If you are reverting to this structure, I would insist we find a way to abstract this custom `Card`.

amonsosanz: Why? it's only 4 repetitions, very localized and the code makes total sense

amonsosanz: Solution agreed on a meet.

PR #267, opened 2023-03-28
feat: adding get connections with credentials

javip97:
This PR enables the possibility of retrieving the credentials of the connection with the query param `credentials`. 
The `query` filter functioning is extended to perform a full text search inside the credential schema attributes.


martinsaporiti (approved):
very good Job!

PR #270, opened 2023-03-28
Pid 472 change link attribute value type

martinsaporiti:


internal/jsonschema/schema.go
```diff
@@ -123,3 +131,81 @@ func (s *JSONSchema) SchemaHash(schemaType string) (core.SchemaHash, error) {
 	id := jsonLdContext + "#" + schemaType
 	return utils.CreateSchemaHash([]byte(id)), nil
 }
+
+// ValidateAndConvert - validates an array of attributes against the schema. Returns a new array with the credential attributes types converted.
+func ValidateAndConvert(ctx context.Context, loader loader.Loader, credentialAttributes []domain.CredentialAttributes) ([]domain.CredentialAttributes, error) {
```
x1m3, line 136: IMHO. This function should be a method of JSONSchema object, so the Loading should be done outside.  

So it could be used like 
remoteSchema, _ := jsonschema.Load(ctx, s.loaderFactory(url))
remoteSchema.ValidateAndConvert()

x1m3 (changes requested):

x1m3 (approved):

PR #271, opened 2023-03-28
Pid 435 issuer can view connections

adrifdez:
Resolves PID 435

To establish a connection, you should use the URL http://localhost:3002/#get-/v1/authentication/qrcode, scan the QR code with the mobile app, expose your app using ngrok, and then replace ISSUER_API_UI_SERVER_URL with your public URL.


ui/src/adapters/api/connections.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { APIResponse, HTTPStatusSuccess, ResultOK, buildAPIError } from "src/adapters/api";
+import { Env } from "src/domain";
+import { API_VERSION, QUERY_SEARCH_PARAM } from "src/utils/constants";
+import { StrictSchema } from "src/utils/types";
+
+const buildAuthorizationHeader = (env: Env) =>
+  `Basic ${window.btoa(`${env.api.username}:${env.api.password}`)}`;
+
+//TODO move it to credentials when is properly cleaned
```
cneuro, line 12: What needs cleaning? Is there an issue associated with this?

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,174 @@
...
+          .map((credential) => (
+            <Typography.Text key={credential.id}>{credential.attributes.type}</Typography.Text>
+          )),
+      title: "Issued credentials",
+    },
+    {
+      dataIndex: "active",
+      key: "active",
+      render: () => <ConnectionsRowDropdown />,
+      width: 55,
```
cneuro, line 57: Why is this needed?

ui/src/adapters/api/connections.ts
```diff
@@ -0,0 +1,87 @@
+import axios from "axios";
+import { z } from "zod";
+
+import { APIResponse, HTTPStatusSuccess, ResultOK, buildAPIError } from "src/adapters/api";
+import { Env } from "src/domain";
+import { API_VERSION, QUERY_SEARCH_PARAM } from "src/utils/constants";
+import { StrictSchema } from "src/utils/types";
+
+const buildAuthorizationHeader = (env: Env) =>
```
cneuro, line 9: This needs to be a helper.

ui/src/adapters/api/connections.ts
```diff
@@ -0,0 +1,87 @@
...
+      headers: {
+        Authorization: buildAuthorizationHeader(env),
+      },
+      method: "GET",
+      params: new URLSearchParams({
+        ...(query !== undefined ? { [QUERY_SEARCH_PARAM]: query } : {}),
+        ...(credentials ? { credentials: "true" } : {}),
+      }),
+      signal,
+      url: `${API_VERSION}/connections`,
```
cneuro, line 53: Any reason the version is here? All other actions have `baseURL: `${env.api.url}/${API_VERSION}`. We should probably provide a helper akin to `buildAPIURL(env)`.

ui/src/adapters/api/connections.ts
adrifdez: We are keeping things there to doesn't break the Credentials section, so I thought to wait until we start that section

ui/src/components/connections/Connections.tsx
adrifdez: Because it's not used in other places

ui/src/adapters/api/connections.ts
cneuro: It's also in schemas.ts, line 9.

ui/src/components/connections/Connections.tsx
cneuro: My question is why do we need it. If we remove it, the columns will scale according to their content, no?

ui/src/adapters/api/connections.ts
cneuro: We will be adding things to credentials.ts anyway, so it won't break anything. Better to add it where it needs to go instead of potentially forgetting it here.

ui/src/adapters/api/connections.ts
adrifdez: I'm following the documentation where the version is related to the path
image

ui/src/components/connections/ConnectionsRowDropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+  {
+    icon: <IconInfoCircle />,
+    key: "details",
+    label: "Details",
+  },
+  {
+    icon: <IconTrash />,
+    key: "delete",
+    label: "Delete connection",
+  },
```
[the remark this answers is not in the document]
amonsosanz: amazing!

ui/src/adapters/api/connections.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface Connection {
+  createdAt: string;
+  credentials: Credential[];
+  id: string;
+  issuerID: string;
+  userID: string;
+}
+
+export async function getConnections({
+  credentials = false,
```
amonsosanz, line 27: I'd set this to `true` by default and let the clients of this function decide when they want to save bandwidth.

adrifdez: I'm going to leave it required to be aware of that field when is used

ui/src/components/connections/Connections.tsx
adrifdez: I'm going to revert to the hardcoded width because I'm seeing that the width is being evenly distributed across each column without it.

ui/src/adapters/api/connections.ts
amonsosanz: Much better!

ui/src/utils/browser.ts
```diff
@@ -1,4 +1,9 @@
 import { APIError, HTTPStatusError } from "src/adapters/api";
+import { Env } from "src/domain";
+
+export function buildAuthorizationHeader(env: Env) {
+  return `Basic ${window.btoa(`${env.api.username}:${env.api.password}`)}`;
+}
```
amonsosanz, line 6: This does not belong here.

I'd move it to `src/adapters/api/index.ts` since it's a utility related to the API authentication.

ui/src/adapters/api/connections.ts
```diff
@@ -0,0 +1,85 @@
...
+        ...(credentials ? { credentials: "true" } : {}),
+      }),
+      signal,
+      url: `${API_VERSION}/connections`,
+    });
+    const { data } = resultOKConnections.parse(response);
+
+    return { data, isSuccessful: true };
+  } catch (error) {
+    return { error: buildAPIError(error), isSuccessful: false };
```
amonsosanz, line 57: This pattern is not correct. We're losing track of the parsing errors. Will open an issue to fix it globally.

ui/src/adapters/api/connections.ts
cneuro: @adrifdez I am referring to what Alex is pointing out, namely to put `API_VERSION` into `baseURL` instead of here in `url`, then to apply this globally via a helper, similar to what you did with `buildAuthorizationHeader()` (calling it e.g. `buildAPIURL()`).

adrifdez: I've solved the problem that Alex mentioned, but I don't see the need to create a function and import it just to add the version.

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,179 @@
...
+      sorter: ({ id: a }, { id: b }) => a.localeCompare(b),
+      title: "Identifier",
+    },
+    {
+      dataIndex: "credentials",
+      ellipsis: { showTitle: false },
+      key: "credentials",
+      render: (credentials: Credential[]) => (
+        <Space>
+          {[...credentials]
```
amonsosanz, line 48: `[...credentials]`?

adrifdez: To perform the sort() operation and keep it immutable.

ui/src/adapters/api/connections.ts
amonsosanz: agreed

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const queryParam = searchParams.get(QUERY_SEARCH_PARAM);
+
+  const tableContents: ColumnsType<Connection> = [
+    {
+      dataIndex: "userID",
+      ellipsis: { showTitle: false },
+      key: "type",
+      render: (userID: Connection["userID"]) => (
+        <Tooltip placement="topLeft" title={userID}>
+          <Typography.Text strong>{userID.replace(/did:\w*:\w*:\w+:/, "")}</Typography.Text>
```
amonsosanz, line 36: Like this, the tooltip shows the prefix and the column does not, which is strange. Also, I'd consider doing this in the persistence layer and storing the result on a different prop, so the UI can widely and consistently use this stripped version of the did.

adrifdez: Probably a function when is used in more places

ui/src/components/connections/Connections.tsx
amonsosanz: Perform the sort() and keep the array immutable? As far as I'm aware we don't have any immutability rule enabled in the linter right? I mean, were you getting an error that made you add this?

I'd love to add [eslint-plugin-functional](https://github.com/eslint-functional/eslint-plugin-functional) to our `@iden3/eslint-config-react-ts` package if that's what you're insinuating! In that case, this would be required.

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,179 @@
...
+    {
+      dataIndex: "credentials",
+      ellipsis: { showTitle: false },
+      key: "credentials",
+      render: (credentials: Credential[]) => (
+        <Space>
+          {[...credentials]
+            .sort((a, b) => a.attributes.type.localeCompare(b.attributes.type))
+            .map((credential) => (
+              <Typography.Text key={credential.id}>{credential.attributes.type}</Typography.Text>
```
amonsosanz, line 51: The design states that the credentials need to be comma separated and this JSX is not doing that. I'd render a single Text and comma-separate the values:

```typescript
<Typography.Text>
  {credentials
    .sort((a, b) => a.attributes.type.localeCompare(b.attributes.type))
    .map((credential) => credential.attributes.type)
    .join(", ")}
</Typography.Text>
```

ui/src/adapters/api/connections.ts
adrifdez: I'm not saying what is best or more appropriate, just keeping align with the way that the documentation explain the endpoints

ui/src/adapters/api/connections.ts
adrifdez: uff I don't know when `oldCredentials` is going to be refactored, so I've renamed and moved the new ones there

ui/src/components/connections/Connections.tsx
amonsosanz: > connection details are going to have different regex and rules

mmm... but then this code will fail or I am missing something?

The user did's I see now have this shape: 

`did:polygonid:polygon:mumbai:2qGGp...Be7kG`

If we can harcode `did:polygonid` we could join the value of the env var `ISSUER_ETHEREUM_RESOLVER_PREFIX` (`polygon:mumbai`) and extract this from the did's. Does it make sense?

And if `did:polygonid` is also going to change, I think we need an env var for it.

This regex does not feel like the way to go.

ui/src/adapters/api/connections.ts
cneuro: Okay, fine then.

ui/src/components/connections/Connections.tsx
amonsosanz: @elias-garcia @cneuro do you agree on enforcing some functional rules and immutability among them?

ui/src/components/connections/Connections.tsx
cneuro: I agree and I think we need more info from API & protocol to come up with a general solution. For now, we should probably just create an issue to track.

adrifdez: `did` is going to be there always, I don't know about polygonid, the others are in env-issuer as the RESOLVER_PREFIX. 
I'm not seeing problems with the regex

ui/src/components/connections/Connections.tsx
adrifdez: image

cneuro: If you mean to say let's consolidate this type of thing, then yes. Otherwise I don't know what you mean by this screenshot without any further context.

adrifdez: I'm not going to create a constant until the moment that it's used in more than one place, don't worry it'll come soon and I'll round it to 60

ui/src/components/connections/Connections.tsx
amonsosanz: Asked [here](https://0xpolygon.slack.com/archives/C0409DDNV7V/p1680249335844219). I wouldn't create a task if it's something we can learn now. I would want to avoid introducing this regex and have the helper function that given a full DID returns the part that we want to present to the user.

amonsosanz: As Jordi commented in the thread the DIDs seem to have a different number of segments (3 according to the [ official docs](https://www.w3.org/TR/did-core/#did-syntax)), but since we're only interested in the last one, we can split it with `:` and show the last segment to the user. This is the way to go.

Also, can we please abstract this to a function and use it both in the table and the tooltip?

ui/src/components/connections/Connections.tsx
elias-garcia: > I'd love to add [eslint-plugin-functional](https://github.com/eslint-functional/eslint-plugin-functional) to our @iden3/eslint-config-react-ts package if that's what you're insinuating! In that case, this would be required.

No, we won't add a full functional package to the iden3 config. I'm okay if you want to add rules related to immutability, like [this](https://github.com/jhusain/eslint-plugin-immutable) plugin

amonsosanz: Works for me. Can you take care of it @elias-garcia?

ui/src/adapters/api/credentials.ts
```diff
@@ -7,21 +7,35 @@ import {
...
+  })
+);
 
 export interface CredentialAttribute {
   attributeKey: string;
   attributeValue: number;
 }
 
-export interface Credential {
+export interface oldCredential {
```
amonsosanz, line 38: I know this is to be removed soon, but it should be uppercased.

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,179 @@
...
+        table={
+          <Table
+            columns={tableContents.map(({ title, ...column }) => ({
+              title: (
+                <Typography.Text type="secondary">
+                  <>{title}</>
+                </Typography.Text>
+              ),
+              ...column,
+            }))}
```
amonsosanz, line 151: Why do we map over the columns to make the title a JSX instead of declaring it as JSX in the first place?

```diff
diff --git a/ui/src/components/connections/Connections.tsx b/ui/src/components/connections/Connections.tsx
index fce23bb..76b1e1e 100644
--- a/ui/src/components/connections/Connections.tsx
+++ b/ui/src/components/connections/Connections.tsx
@@ -38,7 +38,7 @@ export function Connections() {
         </Tooltip>
       ),
       sorter: ({ id: a }, { id: b }) => a.localeCompare(b),
-      title: "Identifier",
+      title: <Typography.Text type="secondary">Identifier</Typography.Text>,
     },
     {
       dataIndex: "credentials",
@@ -52,7 +52,7 @@ export function Connections() {
             .join(", ")}
         </Typography.Text>
       ),
-      title: "Issued credentials",
+      title: <Typography.Text type="secondary">Issued credentials</Typography.Text>,
     },
     {
       dataIndex: "active",
@@ -141,14 +141,7 @@ export function Connections() {
         }
         table={
           <Table
-            columns={tableContents.map(({ title, ...column }) => ({
-              title: (
-                <Typography.Text type="secondary">
-                  <>{title}</>
-                </Typography.Text>
-              ),
-              ...column,
-            }))}
+            columns={tableContents}
             dataSource={connectionsList}
             locale={{
               emptyText:
```

ui/src/components/connections/Connections.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const env = useEnvContext();
+  const [connections, setConnections] = useState<AsyncTask<Connection[], APIError>>({
+    status: "pending",
+  });
+
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const queryParam = searchParams.get(QUERY_SEARCH_PARAM);
+
+  const tableContents: ColumnsType<Connection> = [
```
amonsosanz, line 30: I'd name this `tableColums` since they are columns.

ui/src/components/connections/ConnectionsRowDropdown.tsx
```diff
@@ -0,0 +1,45 @@
...
+    delete: () => void message.error("To develop"),
+    details: () => void message.error("To develop"),
+  };
+
+  const onMenuSelect: MenuProps["onClick"] = ({ domEvent, key }) => {
+    domEvent.stopPropagation();
+    if (key === "details" || key === "delete") {
+      void menuFunction[key]();
+    }
+  };
```
amonsosanz, line 36: What's the practical difference between your implementation and this:

```typescript
export function ConnectionsRowDropdown() {
  const onMenuSelect = () => {
    void message.error("To develop");
  };

  return (
    <Dropdown menu={{ items: MENU_ITEMS, onClick: onMenuSelect }}>
      <Row>
        <IconDots className="icon-secondary" />
      </Row>
    </Dropdown>
  );
}
```

ui/src/components/connections/ConnectionsRowDropdown.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <Dropdown menu={{ items: MENU_ITEMS, onClick: onMenuSelect }}>
+      <Row>
+        <IconDots className="icon-secondary" />
+      </Row>
+    </Dropdown>
+  );
```
amonsosanz, line 44: What's the practical difference between your implementation and this?

```typescript
export function ConnectionsRowDropdown() {
  const onMenuSelect = () => {
    void message.error("To develop");
  };

  return (
    <Dropdown menu={{ items: MENU_ITEMS, onClick: onMenuSelect }}>
      <Row>
        <IconDots className="icon-secondary" />
      </Row>
    </Dropdown>
  );
}
```

adrifdez: That It'll change when we add the different functions

ui/src/components/connections/Connections.tsx
adrifdez: I think that should be to not repeat the JSX for each  title

amonsosanz: It's highly debatable whether doing this map is better than repeating a `Typography.Text` per column.

adrifdez: This is reused everywhere, let me know your preferences and the agreement and I'll change it

amonsosanz: I see, we do it everywhere. Ok, closing this.

amonsosanz (approved):

cneuro (approved):

PR #278, opened 2023-03-30
Consolidated functions, constants and types.

cneuro:
Summary of amends:
- Applying clear & consistent naming to variables & functions
- Ordered utility functions, helpers and exports
- Improved clarity of domain & parsers + no inference
- Parsers are all strict, simplified & named accordingly
- All functions & components are named

There are no functional changes or additions.


ui/src/adapters/parsers/forms.ts
```diff
@@ -143,33 +206,13 @@ const parseIssueCredentialFormDataAttributes = (
...
+}
 
-export const parseLinkExpirationDate = StrictSchema<{
-  linkExpirationDate: dayjs.Dayjs | null;
-}>()(
-  z.object({
-    linkExpirationDate: dayjsInstance.nullable(),
-  })
-);
+// Exports
```
amonsosanz, line 211: I'd remove this comment, there are exported elements above.

ui/src/adapters/parsers/forms.ts
```diff
@@ -143,33 +206,13 @@ const parseIssueCredentialFormDataAttributes = (
...
-    linkMaximumIssuance: z.number().optional(),
-  }),
-});
-
-type IssueCredentialFormDataInput = z.infer<typeof issueCredentialFormDataInput>;
-
-export const issueCredentialFormData = (schemaAttributes: SchemaAttribute[]) => {
-  return StrictSchema<IssueCredentialFormDataInput, CredentialForm>()(
-    issueCredentialFormDataInput.transform(
+export function getIssueCredentialFormData(schemaAttributes: SchemaAttribute[]) {
```
amonsosanz, line 213: This is indeed a parser.

I'd remove the parser `issueCredentialFormDataInputParser` above and integrate it here along with the transformer:

```typescript
export function issueCredentialFormDataParser(schemaAttributes: SchemaAttribute[]) {
  return getStrictParser<IssueCredentialFormDataInput, CredentialForm>()(
    z
      .object({
        attributes: z.object({
          attributes: z.record(z.unknown()),
          expirationDate: dayjsInstanceParser.nullable().optional(),
        }),
        issuanceMethod: z.object({
          linkExpirationDate: dayjsInstanceParser.nullable().optional(),
          linkExpirationTime: dayjsInstanceParser.nullable().optional(),
          linkMaximumIssuance: z.number().optional(),
        }),
      })
      .transform(
        (
          {
            attributes: { attributes, expirationDate },
            issuanceMethod: { linkExpirationDate, linkExpirationTime, linkMaximumIssuance },
          },
          zodRefinementCtx
        ) => {
          const attributesParsingResult = parseIssueCredentialFormDataAttributes(
            attributes,
            schemaAttributes
          );

          if (attributesParsingResult.success) {
            const linkAccessibleUntil = buildLinkAccessibleUntil({
              linkExpirationDate,
              linkExpirationTime,
            });

            const now = new Date().getTime();

            if (linkAccessibleUntil && linkAccessibleUntil.getTime() < now) {
              zodRefinementCtx.addIssue({
                code: z.ZodIssueCode.custom,
                fatal: true,
                message: `${ACCESSIBLE_UNTIL} must be a date/time in the future.`,
              });

              return z.NEVER;
            }

            return {
              attributes: attributesParsingResult.data,
              expiration: expirationDate ? expirationDate.toDate() : undefined,
              linkAccessibleUntil,
              linkMaximumIssuance,
            };
          } else {
            attributesParsingResult.error.issues.forEach(zodRefinementCtx.addIssue);

            return z.NEVER;
          }
        }
      )
  );
}
```

ui/src/domain/index.ts
```diff
@@ -11,160 +11,8 @@ export interface Env {
...
-    };
-  };
-};
-
-export type Schema = Attribute & SchemaProps;
+export type Json = JsonLiteral | { [key: string]: Json } | Json[];
 
 export type JsonLdType = { id: string; name: string };
 
 export type JsonLiteral = string | number | boolean | null;
```
amonsosanz, line 18: `JsonLiteral` is tight to `Json` and should be next to it, ideally above.

ui/src/adapters/parsers/forms.ts
cneuro: Thanks, I'll revise.

ui/src/utils/types.ts
```diff
@@ -20,47 +20,63 @@ interface AsyncTaskPending {
...
+        ? // 3. return type as a ZodSchema parser (it is certain that I/O and parser I/O are identical).
+          ZodSchema<ParserOutput, ZodTypeDef, ParserInput>
+        : // 3. fail
+          never
+      : // 2. fail
+        never
+    : // 1. fail
+      never
+) => ZodSchema<Output, ZodTypeDef, Input> {
+  return (parser: ZodSchema) => parser;
```
amonsosanz, line 81: Nice trick typing the parser just as `ZodSchema` so we can remove the type assertion! Per my tests, this works as well as before. I've tested all the cases from the [original issue](https://github.com/colinhacks/zod/issues/652):

* Omit in a ZodSchema an optional prop of the TS type.
* Forget to make optional in a ZodSchema and optional prop of the TS type.
* Omit in a ZodSchema a required prop of the TS type.
* Make optional in a ZodSchema a required prop of the TS type.
* Add to a ZodSchema a prop that does not exist in the TS type.

amonsosanz: By the way, I yesterday commented that Colin McDonnell published a utility similar to our `getStrictParser`. It's named [toZod](https://github.com/colinhacks/tozod) and sadly, it's not maintained and therefore it's incompatible with the last versions of Zod.

I also commented that it uses [a slightly different version](https://github.com/colinhacks/tozod/blob/master/src/index.ts#L6) of our `Exact` utility type, in case you prefer it:

```typescript
type Exact<X, Y> = [X] extends [Y] ? ([Y] extends [X] ? true : false) : false;
```

ui/src/adapters/api/credentials.ts
```diff
@@ -9,10 +9,10 @@ import {
   buildAPIError,
   buildAuthorizationHeader,
 } from "src/adapters/api";
-import { Schema, schema } from "src/adapters/api/schemas";
+import { SchemaPayload, schemaStrictParser } from "src/adapters/api/schemas";
```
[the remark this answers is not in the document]
cneuro: Thanks again @amonsosanz, this is now implemented as discussed.

ui/src/utils/types.ts
cneuro: Yeah that doesn't make it clearer IMO. At least the current one is shorter.

ui/src/domain/index.ts
```diff
@@ -11,160 +11,8 @@ export interface Env {
   };
```
amonsosanz, line 11: Since we're now following a split domain pattern, I'm tempted to finish the job and create the modules `env.ts` and `json.ts`

ui/src/adapters/api/credentials.ts
```diff
@@ -421,32 +426,36 @@ export type CredentialQRCheck =
...
+const credentialQRCheckParser = getStrictParser<CredentialQRCheck>()(
+  z.union([
+    credentialQRCheckDoneParser,
+    credentialQRCheckErrorParser,
+    credentialQRCheckPendingParser,
+  ])
 );
 
-const resultOKCredentialQRCheck = StrictSchema<ResultOK<CredentialQRCheck>>()(
+const resultOKCredentialQRCheck = getStrictParser<ResultOK<CredentialQRCheck>>()(
```
amonsosanz, line 456: `resultOKCredentialQRCheckParser`

ui/src/domain/index.ts
amonsosanz: The splitting of the domain comes with the drawback of increasing the number of import lines on other modules.

Would you consider maintaining `src/domain/index.ts` and re-export all the inner domains from there?

Like this, we'd keep things organized in different modules and still, we could import every domain entity from `src/domain`.

ui/src/adapters/parsers/index.ts
```diff
@@ -0,0 +1,34 @@
+import { ZodSchema, ZodTypeDef } from "zod";
+
+// The following was implemented due to a perceived limitation of Zod:
+// https://github.com/colinhacks/zod/issues/652
+
+// Asserts whether 2 types are exactly the same.
+type Exact<T, U> = [T, U] extends [U, T] ? true : false;
```
amonsosanz, line 7: I'm ok leaving this here, but be aware that this is a generic utility type, unrelated to the adapters and the parsers, that was better located in `utils/types.ts`.

adrifdez (approved):

ui/src/domain/index.ts
cneuro: Sure, LGTM.

ui/src/adapters/parsers/index.ts
cneuro: But it's only used here. It's directly related to and a part of `getStrictParser`.

ui/src/adapters/parsers/schemas.ts
```diff
@@ -1,349 +1,421 @@
 import { z } from "zod";
 
-import * as domain from "src/domain";
-import { StrictSchema } from "src/utils/types";
+import { getStrictParser } from "src/adapters/parsers";
```
amonsosanz, line 3: I feel some of the changes done in this PR are arbitrary. Specifically in this module, I think the only change I like is adding the `get` prefix to the functions that return a parser. For everything else, I'd revert the changes.

Let me reason a bit about what I don't like:

* In general in the PR you're moving all the types to the top of the module, while we discussed this in the past with the whole team and agreed to keep the types next to where they are used.
* In general in the PR you're adding groups like "// Parsers", "// Composite parsers" 🤯, "// Getters", "// Helpers", "// Exports", etc. IMO this categorization adds poor value and makes things more complex because one element can belong to 2 different categories. Not to mention that Parsers, Composite parsers, and even the Getters are the same to me: just parsers.
* I don't understand this "Composite" prefix you added to some types. Why add it here `type SchemaComposite = AnySchema & SchemaProps` and not here `type ArraySchema = CommonProps & ArrayPropsWithoutItems & { type: "array" }`?. Most types compose and that fact should never be part of its name.
* Same with the parsers. A parser is a parser and apart from the primitives, all parsers compose with others. The fact that we use `foo.and(bar)` to intersect two parsers does not qualify them to be tagged as "Composite parsers".
* Regarding the getters, I don't feel the need for the prefix, but I accept it because it makes more clear that they are function that expects some parameters required by the returned parser. This though does not mean we have to group them all under the "Getters" category.
* Qualifying `unsafelyParseMultiSchemaToDomain` as a helper is 🤯 to me. It's just another parser.
* As commented somewhere else, I'd not add an `// Exports` group since we can have exports in any other group. In this regard, I'd agree to remove all the export to an export statement at the end of the file.

As said, I'd revert all the changes in this module and then add the get prefixes to the functions.

ui/src/domain/index.ts
amonsosanz: Cool. Merged.

ui/src/adapters/parsers/index.ts
amonsosanz: Indeed, it's only used here, but this is a coincidence.

ui/src/components/schemas/ImportSchemaForm.tsx
```diff
@@ -58,21 +58,24 @@ export function ImportSchemaForm({
...
 
   const fetchSchemaFromUrl = (url: string): void => {
-    setSchema({ status: "loading" });
+    setJsonSchema({ status: "loading" });
+
     getSchemaFromUrl({
       url,
     })
-      .then(([schema, rawSchema]) => {
+      .then(([jsonSchema, rawSchema]) => {
```
amonsosanz, line 66: `rawJsonSchema`

ui/src/components/schemas/SchemaDetails.tsx
```diff
@@ -14,23 +14,25 @@ import { ErrorResult } from "src/components/shared/ErrorResult";
...
   const env = useEnvContext();
 
-  const [schemaTuple, setSchemaTuple] = useState<AsyncTask<[Schema, Json], string | z.ZodError>>({
+  const [jsonSchemaTuple, setJsonSchemaTuple] = useState<
+    AsyncTask<[JsonSchema, Json], string | z.ZodError>
+  >({
     status: "pending",
   });
-  const [apiSchema, setApiSchema] = useState<AsyncTask<ApiSchema, APIError>>({
+  const [apiSchema, setApiSchema] = useState<AsyncTask<Schema, APIError>>({
```
amonsosanz, line 35: `schema, setSchema]`

ui/src/components/schemas/SchemaDetails.tsx
```diff
@@ -42,19 +44,21 @@ export function SchemaDetails() {
   const extractError = (error: unknown) =>
     error instanceof z.ZodError ? error : error instanceof Error ? error.message : "Unknown error";
 
-  const fetchSchemaFromUrl = useCallback((apiSchema: ApiSchema): void => {
-    setSchemaTuple({ status: "loading" });
+  const fetchSchemaFromUrl = useCallback((apiSchema: Schema): void => {
```
amonsosanz, line 47: `schema: Schema`

ui/src/domain/jsonSchema.ts
```diff
@@ -0,0 +1,116 @@
+export type Attribute =
+  | StringAttribute
+  | IntegerAttribute
+  | NumberAttribute
+  | BooleanAttribute
+  | NullAttribute
+  | ObjectAttribute
+  | ArrayAttribute
+  | MultiAttribute;
+
```
amonsosanz, line 10: This module was sorted semantically, as the union above ^:

 * First, the primitive values: string, integer, number, and boolean
 * Then, the non-value: null
 * Then, the complex types: object, and array
 * Finally, the composite multi attribute
 
I don't like this new alphabetical sorting that starts with the attribute union and sends the string to the very bottom, below JsonSchema.

We meet on sorting rules and agreed on sorting the types' members and objects' properties, enforcing this with the linter, and we explicitly stated that we were not sorting the code alphabetically. I'd revert this file completely.

cneuro: This is why it's important to add commented sections and to apply some ordering that everyone can see.

cneuro: Revised & improved as discussed.

ui/src/domain/jsonSchema.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export type IntegerAttribute = {
+  name: string;
+  required: boolean;
+  schema: IntegerSchema;
+  type: "integer";
+};
+
+export type IntegerSchema = CommonProps & { enum?: number[]; type: "integer" };
```
amonsosanz, line 28: This should be above `IntegerAttribute`

ui/src/adapters/parsers/jsonSchemas.ts
```diff
@@ -67,96 +114,117 @@ const numberSchemaParser = StrictSchema<domain.NumberSchema>()(
...
-  commonPropsParser.and(
+// Non-primitive
+
+type ArrayPropsInput = {
+  item?: unknown;
+};
+
+type ArraySchemaInput = CommonProps & ArrayPropsInput & { type: "array" };
+
+const arrayPropsParser = getStrictParser<ArrayPropsInput>()(
```
amonsosanz, line 166: Now that these types have the `Input` prefix, their parsers should have it also. The rule is that the name of a parser is formed with the name of the type + the suffix `Parser`.

amonsosanz (approved):

PR #289, opened 2023-04-04
Issuer can delete a connection

adrifdez:
Resolves PID505


ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Typography.Text type="secondary">
+        Identity will be deleted from your connections.
+      </Typography.Text>
+      <Divider />
+      <Space direction="vertical">
+        <Typography.Text strong>Would you also like to:</Typography.Text>
+        <Checkbox
+          onChange={({ target: { checked } }: CheckboxChangeEvent) => setRevokeCredential(checked)}
+        >
+          <Typography.Text>Revoke all credentials for this connections</Typography.Text>
```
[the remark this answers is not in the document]
cneuro: Should also have a period at the end "."

ui/src/components/connections/ConnectionsRowDropdown.tsx
```diff
@@ -1,6 +1,8 @@
 import { Dropdown, Row } from "antd";
+import { useState } from "react";
 import { generatePath, useNavigate } from "react-router-dom";
 
+import { ConnectionDeleteModal } from "./ConnectionDeleteModal";
```
cneuro, line 5: This should be prefaced with the path.

ui/src/components/connections/ConnectionsRowDropdown.tsx
```diff
@@ -26,24 +28,33 @@ const MENU_ITEMS = [
...
-    </Dropdown>
+    <>
+      <Dropdown
+        menu={{
+          items: MENU_ITEMS,
+          onClick: ({ key }) => {
+            onMenuSelect(key);
+          },
+        }}
+        overlayStyle={{ zIndex: 999 }}
```
cneuro, line 51: Now that there is no rendering overlay with the modal, this isn't needed.

ui/src/components/connections/ConnectionsRowDropdown.tsx
cneuro, line 1: Move to Connections, then delete.

cneuro, line 1: Move to ConnectionsTable, then delete.

ui/src/components/connections/Connections.tsx
cneuro, line 1: Rename to `ConnectionsTable`

ui/src/components/connections/ConnectionsRowDropdown.tsx
adrifdez: The dropdown disappears but is still over the modal while hide
dropdownoverlay

cneuro: Yeah that's fine, right?

ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
+import { Checkbox, Divider, Modal, Space, Typography, message } from "antd";
+import { CheckboxChangeEvent } from "antd/es/checkbox";
+import { useState } from "react";
+import { deleteConnection } from "src/adapters/api/connections";
+
+import { ReactComponent as IconClose } from "src/assets/icons/x.svg";
+import { useEnvContext } from "src/contexts/env";
+
+export function ConnectionDeleteModal({
+  callback,
```
cneuro, line 10: This should probably be `onDelete` so it's clear how this prop would be defined.

ui/src/components/connections/ConnectionsRowDropdown.tsx
adrifdez: The modal background should cover everything

ui/src/components/connections/ConnectionsTable.tsx
```diff
@@ -174,6 +216,12 @@ export function Connections() {
           </Row>
         }
       />
+      {connectionSelected && (
+        <ConnectionDeleteModal
```
cneuro, line 220: This (and perhaps the other one too) is missing some sort of refresh for the table, otherwise it may say it's successful but it doesn't show the result.

[Screencast from 05-04-23 12:53:24.webm](https://user-images.githubusercontent.com/2691584/230060638-0a6c2765-4f37-4c58-ba3e-c35b5168d64b.webm)

ui/src/components/connections/ConnectionsTable.tsx
```diff
@@ -220,6 +220,7 @@ export function ConnectionsTable() {
         <ConnectionDeleteModal
           id={connectionSelected}
           onClose={() => setConnectionSelected(undefined)}
+          onDelete={() => void fetchConnections()}
```
cneuro, line 223: These can just be

`onClose={setConnectionSelected}`
`onDelete={fetchConnections}`

?

adrifdez: I should convert `onDelete` to promise

cneuro: So will you do that?

ui/src/components/connections/ConnectionsTable.tsx
```diff
@@ -1,27 +1,35 @@
...
-export function Connections() {
+export function ConnectionsTable() {
   const env = useEnvContext();
+
+  const navigate = useNavigate();
+
   const [connections, setConnections] = useState<AsyncTask<Connection[], APIError>>({
     status: "pending",
   });
+  const [connectionSelected, setConnectionSelected] = useState<string>();
```
amonsosanz, line 32: This stores the ID of a connection to be deleted so I'd name it `[connectionToDelete, setConnectionToDelete]`

ui/src/components/connections/ConnectionsTable.tsx
amonsosanz: We need to pass `undefined` to `setConnectionSelected ` so we can not inline it.

> We should convert `onDelete` to promise

I don't understand this, but I think this code is ok as it's.

adrifdez: I prefer to keep it with just one type, so no, I'll leave it with the void

adrifdez: Both have to be as they are

ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
+import { Checkbox, Divider, Modal, Space, Typography, message } from "antd";
+import { CheckboxChangeEvent } from "antd/es/checkbox";
+import { useState } from "react";
+import { deleteConnection } from "src/adapters/api/connections";
```
amonsosanz, line 4: This belongs to the groups below

ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import { ReactComponent as IconClose } from "src/assets/icons/x.svg";
+import { useEnvContext } from "src/contexts/env";
+
+export function ConnectionDeleteModal({
+  id,
+  onClose,
+  onDelete,
+  open = true,
```
amonsosanz, line 13: I don't think we need this prop. We can pass always true and render the component conditionally.

ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+}: {
+  id: string;
+  onClose: () => void;
+  onDelete?: () => void;
+  open?: boolean;
+}) {
+  const env = useEnvContext();
+  const [revokeCredentials, setRevokeCredential] = useState<boolean>(false);
+  const [deleteCredentials, setRemoveCredential] = useState<boolean>(false);
```
amonsosanz, line 22: `setRevokeCredentials` and `setRemoveCredentials`

ui/src/components/connections/ConnectionDeleteModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const handleDeleteConnection = () => {
+    void deleteConnection({ deleteCredentials, env, id, revokeCredentials }).then((response) => {
+      if (response.isSuccessful) {
+        void message.success(response.data);
```
amonsosanz, line 27: I've noticed @cneuro prefers moving this to the very bottom.

ui/src/components/connections/ConnectionDeleteModal.tsx
adrifdez: `setDeleteCredentials`

ui/src/components/connections/ConnectionDeleteModal.tsx
adrifdez: I thought about it in the beginning but with this approach, we are not mounting and unmounting the component in `ConnectionDetails`

ui/src/components/connections/CredentialsTable.tsx
```diff
@@ -0,0 +1,272 @@
...
+            setCredentials({ error: response.error, status: "failed" });
+          }
+        }
+      }
+    },
+    [userID, env, query, credentialStatus]
+  );
+
+  const handleTypeChange = ({ target: { value } }: RadioChangeEvent) => {
+    const parsedCredentialType = credentialStatusParser.safeParse(value);
```
amonsosanz, line 185: `parsedCredentialStatus`

amonsosanz (approved):

PR #297, opened 2023-04-10
Issuer can submit a credential link issuance

amonsosanz:
Resolves [PID-429](https://polygonid.atlassian.net/browse/PID-429)
Resolves [PID-393](https://polygonid.atlassian.net/browse/PID-393)

Some stuff still pending after this PR is merged:

* implement null attributes (to be defined)
* implement list attributes (to be defined)
* implement multi attributes (to be defined)
* validate the data of the form in the FE
* prevent password managers to autocomplete some form fields (I have an issue with 1password and `id` fields)

Quick demo:

https://user-images.githubusercontent.com/15896177/232119198-dca479f0-c65f-46be-ba5f-eeac35060e15.mp4

[PID-429]: https://polygonid.atlassian.net/browse/PID-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PID-393]: https://polygonid.atlassian.net/browse/PID-393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PID-543]: https://polygonid.atlassian.net/browse/PID-543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/components/credentials/ScanCredentialLink.tsx
```diff
@@ -192,29 +191,27 @@ export function ScanCredentialLink() {
...
 
-              {shareCredentialQRCode.data.offerDetails.attributeValues.map((attribute) => {
-                const formattedValue = formatAttributeValue(
-                  attribute,
-                  //TODO Credentials epic
-                  // shareCredentialQRCode.data.offerDetails.schemaTemplate.attributes
-                  []
-                );
+              {shareCredentialQRCode.data.linkDetail.attributes.map((attribute) => {
+                // const formattedValue = formatAttributeValue(
```
cneuro, line 202: Could you provide the PID for this?

ui/src/styles/index.scss
```diff
@@ -141,6 +141,12 @@
         content: "*";
       }
     }
+    .ant-form-item-control {
+      .ant-form-item-extra {
```
cneuro, line 145: Are these overrides mandated by the designs?

ui/src/components/credentials/SelectSchema.tsx
```diff
@@ -6,11 +6,17 @@ import { getSchemas } from "src/adapters/api/schemas";
...
 import { isAbortedError, makeRequestAbortable } from "src/utils/browser";
 import { SCHEMA_TYPE } from "src/utils/constants";
 
-export function SelectSchema({ schemaID }: { schemaID: string | undefined }) {
+export function SelectSchema({
+  onSelect,
+  schemaID,
+}: {
+  onSelect: (schema: Schema) => void;
+  schemaID: string | undefined;
```
adrifdez, line 18: Now that you have onSelect here we can do `fetchSchemas` in the parent because is weird that the info in the parent comes from the children calling the api

ui/src/components/credentials/Summary.tsx
```diff
@@ -60,139 +34,29 @@ export function Summary({
...
-            href={generatePath(ROUTES.credentialLink.path, { credentialID: credential.id })}
-            icon={<ExternalLinkIcon />}
-            target="_blank"
-            type="link"
-          >
-            Preview credential link
-          </Button>
-        </Row>
+        <Button
+          href={generatePath(ROUTES.credentialLink.path, { linkID })}
```
adrifdez, line 42: You can use `linkURL`

ui/src/domain/jsonSchema.ts
```diff
@@ -45,9 +45,30 @@ export type NumberAttribute = {
   type: "number";
 };
 
+type StringFormat =
```
adrifdez, line 48: Why do we need it if below we accept any string?

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -126,88 +146,101 @@ export function IssueCredential() {
...
+                                const newCredentialFormInput: CredentialFormInput =
+                                  credentialFormInput.issuanceMethod.type === "credentialLink" &&
+                                  updatedValues.expirationDate?.isBefore(
+                                    credentialFormInput.issuanceMethod.linkExpirationDate
+                                  )
+                                    ? {
+                                        ...credentialFormInput,
+                                        issuanceMethod: {
+                                          ...credentialFormInput.issuanceMethod,
+                                          linkExpirationDate: undefined,
```
cneuro, line 206: Why are those needed? Setting them as `undefined` is equivalent to omitting them here.

adrifdez: It's cleaning claimLinkExpiration if the expiration date is before linkExpirationDate. expirationDate has priority over claimLinkExpiration

ui/src/components/credentials/CredentialSubjectForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+          title={shouldShowTitle ? attribute.schema.title || attribute.name : undefined}
+        >
+          {attributeNode}
+        </Card>
+      ) : (
+        <Fragment key={key}>{attributeNode}</Fragment>
+      );
+    });
+
+  return space ? (
```
cneuro, line 102: What does putting a single element in a `Space` do? It only affects the spacing between siblings.

adrifdez: `form` is an array of components.

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -1,122 +1,142 @@
...
         }).then((response) => {
           if (response.isSuccessful) {
-            setCredential({ data: response.data, status: "successful" });
+            setLinkID({ data: response.data.id, status: "successful" });
             setStep("summary");
 
             void message.success("Credential link created");
           } else {
-            setCredential({ error: undefined, status: "failed" });
+            setLinkID({ error: undefined, status: "failed" });
```
cneuro, line 86: Why do we need to make undefined keys `undefined` explicitly?

ui/src/components/credentials/ScanCredentialLink.tsx
amonsosanz: Are these two or just one of them?
* https://polygonid.atlassian.net/browse/PID-431
* https://polygonid.atlassian.net/browse/PID-437

ui/src/styles/index.scss
amonsosanz: The designs are not yet updated with these info messages, [this was recently agreed upon on Slack](https://0xpolygon.slack.com/archives/C03HUC0Q7RU/p1681370779487889). I added these overrides because the default looks ugly IMO:

input-extra-position

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -1,122 +1,142 @@
...
+        ...currentCredentialFormInput,
+        issueCredential: defaultCredentialFormInput.issueCredential,
+      }));
     }
-    return;
-  }, [fetchSchema, schemaID]);
+  }, [schemaID]);
 
   return (
     <SiderLayoutContent
```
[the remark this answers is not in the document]
amonsosanz: Fixed as agreed.

ui/src/components/credentials/CredentialSubjectForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+      const isRootAttribute = parents.length === 0;
+      const shouldShowTitle = isRootAttribute && attribute.type === "object";
+      const key = [...parents, attribute].map((parent) => parent.name).join(" > ");
+
+      return isRootAttribute ? (
+        <Card
+          className="ant-card-type-inner"
+          key={key}
+          title={shouldShowTitle ? attribute.schema.title || attribute.name : undefined}
+        >
```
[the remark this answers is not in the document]
amonsosanz: Awesome! Fixed.

ui/src/components/credentials/ScanCredentialLink.tsx
cneuro: It depends on which part this is - those issues will both change based on what is decided to replace/amend them with. I would reference both here for now.

ui/src/components/credentials/Summary.tsx
amonsosanz: I prefer to keep this relative

ui/src/domain/jsonSchema.ts
amonsosanz: I started creating this as a comment and then moved it to a type but yes, this could be a comment. It's here mainly for documentation purposes so that we know what the standard types are.

ui/src/utils/browser.ts
```diff
@@ -1,22 +1,11 @@
+import { AxiosError } from "axios";
 import { APIError, HTTPStatusError } from "src/adapters/api";
 
-export function downloadFile(blob: Blob, filename: string) {
```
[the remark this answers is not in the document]
amonsosanz: I removed it because I was doing house cleaning and this is not used anywhere.

The JSON is downloaded with `downloadJsonFromUrl` and didn't we agree to remove the download QR button?

ui/src/components/credentials/CredentialSubjectForm.tsx
amonsosanz: Yep `form` is an array of the JSX elements of the cards. Without this space, it looks like this:

image

ui/src/components/credentials/CredentialSubjectForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+  attributes: Attribute[];
+  parents?: ObjectAttribute[];
+  space?: boolean;
+}) {
+  const shouldShowBreadcrumb = useRef<boolean>(true);
+  const form = [...attributes]
+    .sort((a, b) => (a.type !== "object" && b.type !== "object" ? 0 : a.type === "object" ? 1 : -1))
+    .map((attribute: Attribute) => {
+      const showBreadcrumb =
+        attribute.type !== "object" && parents.length > 1 && shouldShowBreadcrumb.current;
```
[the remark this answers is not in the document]
amonsosanz: Did you test the [Family tree](https://schemas.serto.id/schema/family-tree) schema?

ui/src/components/credentials/CredentialSubjectForm.tsx
cneuro: Indeed, but my question is about [why we need the ternary](https://github.com/0xPolygonID/sh-id-platform/pull/297#discussion_r1169788850).

ui/src/components/credentials/IssueCredential.tsx
amonsosanz: The error in `AsyncTaskFailed` is not optional.

ui/src/components/credentials/CredentialSubjectForm.tsx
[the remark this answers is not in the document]
cneuro: My question is related to a less kludgy (more data-driven) way than abusing refs. Doing the following renders it identically unless I'm missing something;

Screenshot from 2023-04-18 12-20-18

ui/src/utils/browser.ts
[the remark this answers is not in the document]
cneuro: > The JSON is downloaded with downloadJsonFromUrl and didn't we agree to remove the download QR button?

Was this discussed and decided on at some point while I was away? I can't find mention of this apart from [this un-answered comment](https://www.figma.com/file/BE3tUbvtSB2EqmJsybpeFC?node-id=5792:165929#419746562), but all the designs remain unchanged.

ui/src/components/credentials/IssueCredential.tsx
cneuro: But here it is bypassing the type checking to essentially become optional. Why are we not passing a string here, at least `response.error.message`? Also the same in [`SelectSchema`](https://github.com/0xPolygonID/sh-id-platform/blob/PID-429-issuer-can-submit-a-credential-link-issuance/ui/src/components/credentials/SelectSchema.tsx#L45).

ui/src/domain/jsonSchema.ts
cneuro: I would add this to the UI README, like we did with the JSON Schema support.

ui/src/components/credentials/ScanCredentialLink.tsx
amonsosanz: Done

adrifdez (approved):

ui/src/domain/jsonSchema.ts
amonsosanz: Yep, I like that! I've also specified that:

> only `date-time`, `time` and `date` will show specialized inputs

ui/src/utils/browser.ts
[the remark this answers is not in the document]
amonsosanz: You will find more context [here](https://0xpolygon.slack.com/archives/C0409DDNV7V/p1681490246248899).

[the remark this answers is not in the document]
cneuro: That's why I am wondering:

> As proposed during the meeting we can include the download QR Code of a credential in the page credential details.

ui/src/components/credentials/CredentialSubjectForm.tsx
amonsosanz: Sorry I did not see your comment, I was adding to Adri's explanation.

We need the ternary because the form is recursive, and per the design, we only want to use cards (and therefore need the space) for the root elements (the first block in the image below).

If you take as an example the schema [Vaccination Certificate](https://schemas.serto.id/schema/vaccination-certificate) and log the `form` variable, you will see that the component `CredentialSubjectForm` is called 4 times, the first one is the one called from the outside (from `IssueCredentialForm`) but then, there are 3 more calls which occur recursively, from the component itself, to render the fields of the 3 objects present in the schema:

image
 
In the first call, we render the root elements with cards and therefore need the spaces. Later though, in the recursive calls, we don't want to space the input fields that are rendered inside the cards.

I have though realized that we don't need the prop `space` since we can reuse the variable `isRootAttribute` to infer the need for a space in the ternary. I've pushed this improvement.

ui/src/utils/browser.ts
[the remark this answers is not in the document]
amonsosanz: Yes, if we finally decide to do that, we will have to add again this function.

ui/src/components/credentials/CredentialSubjectForm.tsx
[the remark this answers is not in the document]
amonsosanz: My initial attempt was to use the condition `index === 0` and it was not working but I don't recall the reason. It was surely with a previous implementation because now it works as you point out.

Love removing this ref! Done!

ui/src/components/credentials/IssueCredential.tsx
cneuro: Amend pushed as discussed.

cneuro (approved):

PR #301, opened 2023-04-11
feat: adding send push notification with pubsub

javip97:


internal/core/domain/claim.go
```diff
@@ -36,10 +36,10 @@ type Claim struct {
 	CoreClaim        CoreClaim       `json:"core_claim"`
 	MTPProof         pgtype.JSONB    `json:"mtp_proof"`
 	SignatureProof   pgtype.JSONB    `json:"signature_proof"`
-	IdentityState    *string         `json:"-"`
+	IdentityState    *string         `json:"identity_state"`
```
x1m3, line 39: Probably we do not need to export it as json now, as we are only adding the ID of the claim and not all the data.

internal/core/domain/claim.go
```diff
@@ -36,10 +36,10 @@ type Claim struct {
...
 	Status           *IdentityStatus `json:"status"`
 	CredentialStatus pgtype.JSONB    `json:"credential_status"`
-	HIndex           string          `json:"-"`
+	HIndex           string          `json:"h-index"`
```
x1m3, line 42: Same as identityState. Probably not needed.

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,50 @@
...
+	Publish(ctx context.Context, topic string, payload Event) error
+}
+
+// EventHandler is the type that functions that handle an event must comply.
+type EventHandler func(context.Context, Event) error
+
+// Subscriber subscribes to the pubsub topics
+type Subscriber interface {
+	Subscribe(ctx context.Context, topic string, callback EventHandler)
+	Unsubscribe(ctx context.Context, topic string) error
```
x1m3, line 43: I think we can remove Unsuscribe for now. Not used and not clear how to use it.

pkg/pubsub/redis.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/go-redis/redis/v8"
+
+	"github.com/polygonid/sh-id-platform/internal/log"
+)
+
+// RedisClient struct
+type RedisClient struct {
+	conn        *redis.Client
+	subscribers map[string]*redis.PubSub
```
x1m3, line 16: Probably not needed if we remove the Unsubscribe method

internal/core/services/notification.go
```diff
@@ -0,0 +1,107 @@
...
+		return errors.New("sendCredentialNotification unexpected data type")
+	}
+
+	return n.sendCreateCredentialNotification(ctx, cEvent.ID)
+}
+
+func (n *notification) sendCreateCredentialNotification(ctx context.Context, credID string) error {
+	credUUID, err := uuid.Parse(credID)
+	if err != nil {
+		return err
```
x1m3, line 48: Log the credentiaID, etc... Any info to understand what is happening.

internal/core/domain/claim.go
javip97: agree, I added them as I was sending the credential in the pubsub before

javip97: agree, I added them as I was sending the credential in the pubsub before

internal/core/services/notification.go
javip97: sure, thanks!

pkg/pubsub/redis.go
x1m3, line 1: We should write a small test here doing a publish and a subcribe to be sure that we receive the same message we send.

x1m3 (changes requested):

pkg/pubsub/redis.go
```diff
@@ -0,0 +1,57 @@
...
+// Publish publishes a new topic payload
+func (rdb *RedisClient) Publish(ctx context.Context, topic string, payload Event) error {
+	return rdb.conn.Publish(ctx, topic, payload).Err()
+}
+
+// Subscribe adds a topic to the
+func (rdb *RedisClient) Subscribe(ctx context.Context, topic string, callback EventHandler) {
+	pubsub := rdb.conn.Subscribe(ctx, topic)
+
+	go func() {
```
x1m3, line 31: Listen for context cancellation, run pubsub.Close and quit the loop

x1m3 (changes requested):

pkg/pubsub/redis.go
x1m3: We agreed on doing it in another feature.

x1m3 (approved):

x1m3 (approved):

martinsaporiti (approved):
LGTM

PR #306, opened 2023-04-13
Do not throw CreateCredential event for mtProof credentials

x1m3:
We will send this notification when it is confirmed that is saved in the blockchain


javip97 (approved):
LGTM

martinsaporiti (approved):
LGTM

PR #308, opened 2023-04-13
feat: sending push notification for new connections

javip97:
This PR enables sending push notifications with the previously created credentials to authenticated users.

Use case: 
1. An issuer direct issue a credential to a user.
2. The user authenticates
3. The user receives a push notification with the credential offer


internal/core/services/identity.go
```diff
@@ -63,10 +64,11 @@ type identity struct {
 
 	ignoreRHSErrors bool
 	rhsPublisher    reverse_hash.RhsPublisher
+	pubsub          pubsub.Client
```
x1m3, line 67:
```suggestion
	pubsub          pubsub.Publisher
```
Just a suggestion. Using pubsub.Client is perfectly acceptable. But injecting pubsub.Publisher tells the reader that this service is publishing messages but not listening to them.

x1m3 (approved):
LGTM

martinsaporiti (approved):
Great!

PR #311, opened 2023-04-14
fix: change state to failed when publisher fails

martinsaporiti:


x1m3 (approved):

x1m3 (commented):
GOOOD Catch!!!

javip97 (approved):
Nice one

PR #312, opened 2023-04-14
Test and small changes in pubsub 

x1m3:
- [x] Added a basic test
- [x] Created a wrapper object to encapsulate messages
- [x] Distinguish between Events and Messages.
- [x] Events are responsible of serialize to / unserialize from Messages
- [x] Added advanced test
- [x] Do not depend of internal log
- [x] Handle panics in callbacks to avoid pubsub crashing


javip97 (approved):

internal/core/event/event.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/polygonid/sh-id-platform/pkg/pubsub"
+)
+
+const (
+	CreateCredentialEvent = "createCredentialEvent" // CreateCredentialEvent create credential event
+	CreateConnectionEvent = "createConnectionEvent" // CreateConnectionEvent create connection MyEvent
```
martinsaporiti, line 11: should be event?

internal/core/event/event.go
martinsaporiti, line 1: should this file be placed inside the core?

martinsaporiti (approved):
LGTM, I left two comments.

internal/core/event/event.go
x1m3: It could be ID or Topic? Probably the best could be just CreateCredential (without the event word)  but then it collides with the struct name.

internal/core/event/event.go
x1m3: This are domain events. I find it acceptable to be near the domain core entities, service, etc..But no strong preferences.

PR #320, opened 2023-04-18
Flow to delete link

adrifdez:


ui/src/components/credentials/LinkDeleteModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+      maskClosable
+      okButtonProps={{ danger: true }}
+      okText="Delete"
+      onCancel={onClose}
+      onOk={handleDeleteLink}
+      open
+      title="Are you sure you want to delete this credential link?"
+    >
+      <Typography.Text type="secondary">
+        Users will not be able to receive this credential anymore. This action cannot be undone.
```
cneuro, line 46: Small typo: "anymore" should be "any more" (designs are wrong too), in fact it should actually say "any longer", since that is grammatically correct.

amonsosanz: are you sure?

image

ui/src/adapters/api/credentials.ts
```diff
@@ -149,7 +149,7 @@ export async function linkUpdate({
   };
 }): Promise<APIResponse<string>> {
   try {
-    const response = await axios<{ message: string }>({
+    const response = await axios({
```
amonsosanz, line 152: Wasn't this fixed in the other PR? why is the fix also here?

ui/src/components/credentials/LinkDeleteModal.tsx
cneuro: Indeed, you are right. "Anymore" refers to time, and "any more" to quantities according to [Grammarly](https://www.grammarly.com/blog/anymore-vs-any-more/).

Nevertheless, it seems that "any longer" is [more correct here](https://learningenglish.voanews.com/a/anymore-any-longer-and-no-longer/6315945.html).

ui/src/adapters/api/credentials.ts
adrifdez: This one comes from develop not revoke credentials

cneuro (approved):

ui/src/adapters/api/credentials.ts
amonsosanz: Of course, my question is: why do u fix this twice?

ui/src/adapters/api/credentials.ts
```diff
@@ -143,21 +143,13 @@ export const linkStatusParser = getStrictParser<LinkStatus>()(
   z.union([z.literal("active"), z.literal("inactive"), z.literal("exceeded")])
 );
 
-const linkAttributes = getStrictParser<LinkAttributes>()(
```
amonsosanz, line 146: Why are you removing the Link attributes in this PR?

adrifdez: Because it's not used and has changed in the api

amonsosanz: Adri updated it's local env and was getting parsing errors in the links list because the `attributes` prop is now gone.

ui/src/adapters/api/credentials.ts
adrifdez: To ensure that I don't forget to import the parser when having a conflict at the merge

amonsosanz (approved):

PR #322, opened 2023-04-18
New section Issuer state

adrifdez:
Resolves PID-569 PID-570 PID-572 PID-573 PID-574

To be able to publish a new state you should create a credential directly in the API or revoke it in the UI

You should have a positive balance in your wallet associated with your issuer


ui/src/domain/issuer-state.ts
cneuro, line 1: This should be imported & exported in `domain/index.ts`.

ui/src/components/shared/CredentialRevokeModal.tsx
```diff
@@ -29,18 +31,7 @@ export function CredentialRevokeModal({
         onClose();
         onRevoke();
```
cneuro, line 32: This is missing (here or inside the callback itself) the additional notification that the credential has been revoked. See [comment](https://www.figma.com/file/BE3tUbvtSB2EqmJsybpeFC?node-id=5808:184164#417295665).

ui/src/components/shared/TableCard.tsx
```diff
@@ -18,16 +18,16 @@ export function TableCard({
...
+  query?: string | null;
+  searchPlaceholder?: string;
   showDefaultContents: boolean;
   table: ReactNode;
   title: ReactNode;
 }) {
   return (
     <Card bodyStyle={{ padding: 0 }} title={title}>
-      {!showDefaultContents && (
+      {!showDefaultContents && onSearch && searchPlaceholder && query !== undefined && (
```
cneuro, line 30: If `query` is defaulted to `null`, then checking for `query !== undefined` doesn't make sense?

ui/src/contexts/issuer-state.tsx
```diff
@@ -0,0 +1,79 @@
...
+        if (!isAbortedError(response.error)) {
+          void message.error(response.error.message);
+        }
+      }
+    },
+    [env]
+  );
+
+  const notifyChange = useCallback(() => {
+    void message.success({
```
cneuro, line 50: I strongly believe this should be `message.info`. Let's [clarify](https://www.figma.com/file/BE3tUbvtSB2EqmJsybpeFC?node-id=5808:184164#417295665) this with Liudas.

ui/src/components/shared/CredentialRevokeModal.tsx
adrifdez: Sorry, but I don't get what it's missing, you are pointing to a line without changes and the notification is below that line

.env-ui.sample
```diff
@@ -1,2 +1,3 @@
+BLOCK_EXPLORER_URL=https://mumbai.polygonscan.com
```
cneuro, line 1: This and the local UI dev equivalent needs to be documented in the root README.

ui/src/adapters/api/issuer-state.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+export async function getStatus({
+  env,
+  signal,
+}: {
+  env: Env;
+  signal?: AbortSignal;
+}): Promise<APIResponse<boolean>> {
```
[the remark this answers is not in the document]
adrifdez: Could we call it `pendingActions` to be resued in the parser?

ui/src/adapters/api/issuer-state.ts
```diff
@@ -0,0 +1,131 @@
...
+      url: `${API_VERSION}/state/publish`,
+    });
+
+    return { data: true, isSuccessful: true };
+  } catch (error) {
+    return { error: buildAPIError(error), isSuccessful: false };
+  }
+}
+
+export async function retryState({ env }: { env: Env }): Promise<APIResponse<boolean>> {
```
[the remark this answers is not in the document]
adrifdez: retryPublishState

ui/src/components/shared/CredentialRevokeModal.tsx
cneuro: We need a `message.success` that displays "Credential successfully revoked" (or whatever the BE returns after revocation) in addition to the issuer state notification.

adrifdez: I think that is more related to the decision that you'll  reach in figma, if we change it to an info message it'll has sense but if not I think that we can leave with just this one

cneuro: Just for consistency though, we should always return the operation's message regardless of the issuer state notification.

ui/src/adapters/api/issuer-state.ts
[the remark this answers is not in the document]
amonsosanz: ok let's keep the name `pendingActions`, I guess it makes sense if the API also uses this name.

ui/src/components/shared/CredentialRevokeModal.tsx
```diff
@@ -3,8 +3,8 @@ import { useState } from "react";
 
 import { revokeCredential } from "src/adapters/api/credentials";
```
amonsosanz, line 4: Can you please update the literal:

> Revoking of a credential must be accompanied by publishing of issuer state in order for  the action to be recorded

as:

> Revoking of a credential must be accompanied by publishing of issuer state in order for  the action to be effective

To be consistent with the delete connection msg?

ui/src/components/shared/CredentialRevokeModal.tsx
```diff
@@ -3,7 +3,8 @@ import { useState } from "react";
```
amonsosanz, line 3: Can you please update the literal:

> Revoking of a credential must be accompanied by publishing of issuer state in order for  the action to be recorded

as:

> Revoking of a credential must be accompanied by publishing of issuer state in order for  the action to be effective

To be consistent with the delete connection msg?

ui/src/contexts/IssuerState.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, [refreshStatus]);
+
+  const value = useMemo(() => {
+    return { notifyChange, refreshStatus, status };
+  }, [notifyChange, refreshStatus, status]);
+
+  return <IssuerStateContext.Provider value={value} {...props} />;
+}
+
+export function useStateContext() {
```
amonsosanz, line 77: `useIssuerStateContext`

.env-ui.sample
cneuro: This also needs to be `ISSUER_UI_BLOCK_EXPLORER_URL` to remain consistent with the rest of the env vars.

ui/src/contexts/issuer-state.tsx
cneuro: Hi @Figmintojas could you quickly weigh in here? Basically, instead of this:

Screenshot from 2023-04-20 12-19-48

I am advocating to [keep messaging consistent](https://www.figma.com/file/BE3tUbvtSB2EqmJsybpeFC/Polygon-ID-Platform?node-id=5806-174811&t=92KxN4W0RYdvJLdK-0) by always accompany it with the operational feedback and changing the icon to represent the message type. Like this:

Screenshot from 2023-04-20 12-21-29

ui/src/components/issuer-state/IssuerState.tsx
```diff
@@ -0,0 +1,266 @@
...
+        clearInterval(checkUnpublished);
+      }
+    }, POLLING_INTERVAL);
+
+    return () => clearInterval(checkUnpublished);
+  }, [fetchTransactions, transactionsList]);
+
+  return (
+    <SiderLayoutContent
+      description="Issuing merkle tree type credentials and revoking credentials require an additional step, know as publishing issuer state."
```
cneuro, line 194: "Merkle" should be [capitalized](https://en.wikipedia.org/wiki/Merkle_tree).

ui/src/contexts/issuer-state.tsx
amonsosanz: I think we can go ahead with this. It's pretty obvious you're right @cneuro. The first msg is info, the second is a success op confirmation.

README.md
```diff
@@ -125,12 +126,12 @@ Completing either option of the [installation](#installation) process yields the
...
-    - `VITE_API_PASSWORD -> ISSUER_API_UI_AUTH_PASSWORD`
-    - `VITE_ISSUER_DID -> ISSUER_API_UI_ISSUER_DID`
-    - `VITE_ISSUER_NAME -> ISSUER_API_UI_ISSUER_NAME`
-    - `VITE_ISSUER_LOGO -> ISSUER_API_UI_ISSUER_LOGO`
+   - `VITE_API_URL -> ISSUER_API_UI_SERVER_URL`
+   - `VITE_API_USERNAME -> ISSUER_API_UI_AUTH_USER`
+   - `VITE_API_PASSWORD -> ISSUER_API_UI_AUTH_PASSWORD`
+   - `VITE_ISSUER_DID -> ISSUER_API_UI_ISSUER_DID`
+   - `VITE_ISSUER_NAME -> ISSUER_API_UI_ISSUER_NAME`
+   - `VITE_ISSUER_LOGO -> ISSUER_API_UI_ISSUER_LOGO`
```
amonsosanz, line 134: isn't `VITE_BLOCK_EXPLORER_URL` missing here?

ui/src/components/issuer-state/IssuerState.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const [isPublishing, setIsPublishing] = useState<boolean>(false);
+  const [isRetrying, setIsRetrying] = useState<boolean>(false);
+  const [transactions, setTransactions] = useState<AsyncTask<Transaction[], APIError>>({
+    status: "pending",
+  });
+
+  const transactionsList = useMemo(
+    () => (isAsyncTaskDataAvailable(transactions) ? transactions.data : []),
+    [transactions]
+  );
```
amonsosanz, line 47: Why did you need to use `useMemo` in this case?

ui/src/contexts/IssuerState.tsx
```diff
@@ -0,0 +1,79 @@
...
+      } else {
+        if (!isAbortedError(response.error)) {
+          void message.error(response.error.message);
+        }
+      }
+    },
+    [env]
+  );
+
+  const notifyChange = useCallback(() => {
```
amonsosanz, line 49: Currently, this does not inform the user about any type of pending to-be-published change. It's very specific to revocation changes. If this is expected to remain specific, I'd rename it as `notifyPendingRevocationChange`.

I suspect though that this msg will be also required to be shown when we issue new direct credentials. In such a case, I'd keep the name as it's now but require the msg `Revocation requires issuer state to be published` to be passed as a param.

ui/src/adapters/api/issuer-state.ts
```diff
@@ -0,0 +1,131 @@
...
+    const response = await axios({
+      baseURL: env.api.url,
+      headers: {
+        Authorization: buildAuthorizationHeader(env),
+      },
+      method: "GET",
+      signal,
+      url: `${API_VERSION}/state/transactions`,
+    });
+    const { data } = resultOKTransactionsParser.parse(response);
```
amonsosanz, line 118: I'd sort the tx's so the newer ones are at the top, as we do with the schemas.

ui/src/components/issuer-state/IssuerState.tsx
```diff
@@ -0,0 +1,266 @@
...
+      }
+
+      void refreshStatus();
+      void fetchTransactions();
+
+      setIsPublishing(false);
+    });
+  };
+
+  const retry = () => {
```
cneuro, line 69: This is almost identical to `publish` ... can we DRY this up? Maybe at least with the success message?

ui/src/contexts/IssuerState.tsx
adrifdez: I'd remove this and reuse `refreshStatus` then

cneuro (approved):

amonsosanz (approved):

PR #323, opened 2023-04-18
Pid 567 send notification when mpt credentials are published

martinsaporiti:
* This PR has the code for sending push notifications when the state is published. :rocket:
* Push notifications are sent only when the credential (the link indeed) has mtp proof as not null :helicopter:


javip97 (approved):
LGTM

internal/repositories/claims.go
```diff
@@ -876,6 +876,67 @@ func (c *claims) GetClaimsIssuedForUser(ctx context.Context, conn db.Querier, id
...
+		FROM claims
+		WHERE claims.identifier = $1
+		AND identity_state = $2
+		AND mtp_proof IS NOT NULL
+	`
+	rows, err := conn.Query(ctx, query, did.String(), state)
+	if err != nil {
+		return nil, err
+	}
+
```
x1m3, line 907: defer rows.Close()

x1m3 (changes requested):

internal/repositories/claims.go
```diff
@@ -876,6 +876,67 @@ func (c *claims) GetClaimsIssuedForUser(ctx context.Context, conn db.Querier, id
...
+			&claim.CredentialStatus,
+			&claim.Revoked,
+			&claim.CoreClaim)
+		if err != nil {
+			return nil, err
+		}
+		claims = append(claims, &claim)
+	}
+
+	defer rows.Close()
```
x1m3, line 936: This should go up

x1m3 (changes requested):
There is a defer rows.Close that should go up in the code

x1m3 (approved):
Nice.

PR #334, opened 2023-04-22
Issuer can view details of issued credential links

amonsosanz:
Resolves PID-562


ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -0,0 +1,248 @@
...
+              <Space direction="vertical" size="large">
+                <Card className="background-grey">
+                  <Space direction="vertical">
+                    <Typography.Text type="secondary">CREDENTIAL LINK DETAILS</Typography.Text>
+
+                    <Detail flavor={flavor} label="Link status" text={text} />
+
+                    <Detail label="Proof type" text={proofTypes.join(", ")} />
+
+                    <Detail label="Creation date" text="-" />
```
amonsosanz, line 206: Backend is adding this field

amonsosanz: Done

ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+                    <Detail
+                      label="Credential expiration date"
+                      text={expiration ? formatDate(expiration, "date-time") : "-"}
+                    />
+
+                    <Detail copyable label="Schema hash" text="-" />
```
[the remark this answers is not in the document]
amonsosanz: Done

ui/src/adapters/api/issuer-state.ts
```diff
@@ -26,7 +26,7 @@ const transactionStatusParser = getStrictParser<TransactionStatus>()(
 const transactionParser = getStrictParser<Transaction>()(
   z.object({
     id: z.number(),
-    publishDate: z.coerce.date(),
+    publishDate: z.coerce.date(z.string().datetime()),
```
cneuro, line 29: Why do we get an error if we just do `z.string().datetime()` without the `coerce`? I don't feel courageous enough to `infer` the cathedral ...

ui/src/components/credentials/ObjectAttributeValueTreeNode.tsx
```diff
@@ -0,0 +1,90 @@
...
+              : attributeValue.value;
+          }
+          default: {
+            return attributeValue.value === undefined ? "-" : attributeValue.value;
+          }
+        }
+      }
+    }
+    case "object": {
+      return undefined;
```
cneuro, line 38: This shouldn't be `null`?

ui/src/utils/forms.ts
```diff
@@ -1,7 +1,11 @@
 import dayjs from "dayjs";
 
-export function formatDate(date: dayjs.Dayjs | Date, showTime?: boolean) {
-  const format = showTime ? "YYYY-MM-DD HH:mm" : "YYYY-MM-DD";
+export function formatDate(
+  date: dayjs.Dayjs | Date,
+  format: "date" | "date-time" | "time" = "date"
```
cneuro, line 5: [Possibly](https://github.com/0xPolygonID/sh-id-platform/pull/334/files#r1176223171) should be `format: "date" | "date-time" | "time" = "date-time"`

ui/tsconfig.json
```diff
@@ -2,9 +2,9 @@
   "compilerOptions": {
     "allowJs": false,
     "allowSyntheticDefaultImports": true,
-    "alwaysStrict": true,
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
+    "alwaysStrict": true,
```
cneuro, line 7: Alex, you are going to the dark side. I like it.

ui/src/components/credentials/ObjectAttributeValueTreeNode.tsx
```diff
@@ -0,0 +1,90 @@
...
+            {attributeValue.type === "object" ? (
+              <IconTreeNode className="icon-secondary" />
+            ) : (
+              <IconTreeLeaf className="icon-secondary" />
+            )}
+            <Typography.Text>{name}</Typography.Text>
+          </Space>
+        </Col>
+
+        <Col style={{ marginLeft: 28 }}>
```
cneuro, line 76: `marginLeft: 28` override is hardcoded in lots places. Can we DRY this up?

ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+              }
+            }
+          })();
+
+          return (
+            <Card
+              className="centered"
+              extra={
+                <Button danger icon={<IconTrash />} onClick={() => setShowModal(true)} type="text">
+                  Delete Link
```
cneuro, line 194: We don't capitalize "link" or "credential" or "schema" etc. anywhere else so we shouldn't start here.

ui/src/adapters/api/issuer-state.ts
amonsosanz: Because `z.string().datetime()` first parses a string and then makes sure it has the expected datetime format, leaving the value untouched as a string.

Our type though dictates that `publishDate` must be a `Date`, so after we make sure the string is a valid datetime, we have to coerce it (convert it) to a `Date`.

ui/src/components/credentials/ObjectAttributeValueTreeNode.tsx
amonsosanz: No, because we have an attribute whose type is `null`. This `undefined` represents the absence of a value, and we need to differentiate it from the legit value `null` and the non-provided value for optional attributes which is represented by the string literal `"-"`.

ui/src/components/credentials/ObjectAttributeValueTreeNode.tsx
```diff
@@ -0,0 +1,90 @@
...
+      if (value === undefined) {
+        return "-";
+      } else {
+        switch (attributeValue.schema.format) {
+          case "date":
+          case "date-time":
+          case "time": {
+            const parsedDate = z.coerce.date(z.string().datetime()).safeParse(value);
+            return parsedDate.success
+              ? formatDate(parsedDate.data, attributeValue.schema.format)
```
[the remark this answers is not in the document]
amonsosanz: The second parameter is already optional and it defaults to `date`. Am I missing something?

ui/src/utils/forms.ts
amonsosanz: Okay, so we definitely want to display the time in almost all cases.

amonsosanz: Done!

ui/src/components/shared/Detail.tsx
```diff
@@ -1,19 +1,31 @@
...
   copyable?: boolean;
   ellipsisPosition?: number;
+  flavor?:
+    | {
+        type: "plain";
+      }
+    | {
+        color?: string;
+        type: "tag";
+      };
```
adrifdez, line 24: Instead of flavor just tag
```suggestion
  tag?: PresetStatusColorType
```

ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+        (jsonSchema.type === "object" &&
+          jsonSchema.schema.properties
+            ?.filter((child): child is ObjectAttribute => child.type === "object")
+            .find((child) => child.name === "credentialSubject")) ||
+        null;
+
+      if (credentialSubject) {
+        const parsedCredentialSubject = getAttributeValueParser(credentialSubject).safeParse({
+          ...link.credentialSubject,
+          // Add required id property that links lack
```
cneuro, line 62: Is this a BE issue? Either way we need to create one and reference it here.

ui/src/components/schemas/ImportSchemaForm.tsx
```diff
@@ -54,13 +54,10 @@ export function ImportSchemaForm({
         }
   );
 
-  const processError = (error: unknown) =>
-    error instanceof z.ZodError ? error : error instanceof Error ? error.message : "Unknown error";
-
-  const fetchSchemaFromUrl = (url: string): void => {
+  const fetchJsonSchemaFromUrl = (url: string): void => {
```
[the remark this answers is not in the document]
amonsosanz: No, but it's always safer to be explicit about the types. I've realized you have removed several explicit return types from functions in some refactors and it's not a good idea IMO. I'd go the opposite way: Add a linter rule to require all functions to declare their return type.

ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+          id: "",
```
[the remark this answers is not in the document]
amonsosanz: why should we?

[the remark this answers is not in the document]
amonsosanz: Prop removed.

ui/src/components/shared/Detail.tsx
amonsosanz: What is `PresetStatusColorType`?

When you propose `tag`, I assume you are not referring to a UX "tag" but to a kind of `_tag` discriminator, right?

ui/src/components/credentials/LinkDetails.tsx
amonsosanz: This has changed in the last review of the code [per a suggestion](#discussion_r1176430874) from @adrifdez. Now instead of artificially adding a value, we're artificially removing the `id` attribute from the JSON Schema, which apart from being more correct, has the beneficial side effect of removing the attribute from the tree.

We do not have to create any issues for this because there is nothing to solve or fix. This is expected to be like this. The reason is that we are using the JSON Schema to create the form and the tree representation of the credential Links and this has the side effect that the required field "id" (which will hold the DID) can't be yet provided at this stage of the flow, since it is yet unknown to whom the credential will be directed. Since it's required schema-wise, we have to artificially remove it to prevent validation errors. As you can see [here](https://github.com/0xPolygonID/sh-id-platform/blob/develop/ui/src/components/credentials/IssueCredentialForm.tsx#L36-L42) it's also removed from the form when creating the links.

ui/src/components/credentials/ObjectAttributeValueTreeNode.tsx
cneuro: A component is probably not worth it, but adding a constant is.

ui/src/components/schemas/ImportSchemaForm.tsx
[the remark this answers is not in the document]
cneuro: But if it's correctly inferred, why bother? Either way, I've added an [issue](https://polygonid.atlassian.net/browse/PID-637) to discuss this.

ui/src/components/shared/Detail.tsx
amonsosanz: Done as agreed. Now instead of `flavor` we can optionally pass a `TagProps` object to render the detail as a tag.

cneuro (approved):

adrifdez (approved):

PR #336, opened 2023-04-24
feat: adding get credential qr code endpoint in api_ui

javip97:
This endpoint enables generating a qr code of a credential so the user can fetch it through a QR Code


martinsaporiti (approved):
LGTM

x1m3 (approved):

PR #338, opened 2023-04-24
chore: send only one push notif. for all the pending credentials

martinsaporiti:
In this PR we improve the push notification in order to sen only one for all the pending credentials when the state is published.


internal/gateways/publisher.go
```diff
@@ -432,13 +432,24 @@ func (p *publisher) updateIdentityStateTxStatus(ctx context.Context, state *doma
...
+		grupedCredentials := groupByUserId(claimsToNotify)
+		for _, claims := range grupedCredentials {
+			err = p.notificationPublisher.Publish(ctx, event.CreateCredentialEvent, &event.CreateCredential{CredentialIDs: claims, IssuerID: state.Identifier})
 			if err != nil {
-				log.Error(ctx, "publish EventCreateCredential", "err", err.Error(), "credential", c.ID.String())
+				log.Error(ctx, "publish EventCreateCredential", "err", err.Error(), "credential", claims)
 				continue
 			}
 		}
+
```
javip97, line 445: let's remove this comments if that block is not used

x1m3, line 445: Remove comments

internal/gateways/publisher.go
```diff
@@ -432,13 +432,24 @@ func (p *publisher) updateIdentityStateTxStatus(ctx context.Context, state *doma
 			return err
 		}
 		log.Info(ctx, "sending notifications:", "numberOfClaims", len(claimsToNotify))
-		for _, c := range claimsToNotify {
-			err = p.notificationPublisher.Publish(ctx, event.CreateCredentialEvent, &event.CreateCredential{CredentialID: c.ID.String(), IssuerID: state.Identifier})
+
+		// TODO - group by userID
```
x1m3, line 436: TODO?

internal/gateways/publisher.go
```diff
@@ -452,6 +463,15 @@ func (p *publisher) updateIdentityStateTxStatus(ctx context.Context, state *doma
 	return nil
 }
 
+// groupByUserId - groups claims by user id
+func groupByUserId(claims []*domain.Claim) map[string][]string {
```
x1m3, line 467: if makes more sense to me returning a  map[string][]domain.Claims. Just cosmetic but in it explains better the purpose of the function

x1m3 (commented):
Just left some comment with minor suggestions

internal/core/services/tests/notification_test.go
x1m3, line 1: Why removing this test?

internal/core/services/notification.go
```diff
@@ -51,44 +51,77 @@ func (n *notification) SendCreateConnectionNotification(ctx context.Context, e p
...
-		return err
-	}
+	currentUserID := ""
+	var currentConnection *domain.Connection
+	credentials := make([]*domain.Claim, 0)
+	for i, credID := range credIDs {
+		credUUID, err := uuid.Parse(credID)
+		if err != nil {
+			log.Error(ctx, "sendCreateCredentialNotification: failed to parse credID", "err", err.Error(), "issuerID", issuerID, "credID", credID)
+			continue
```
x1m3, line 68: I think it could be better to return error like before

internal/gateways/publisher.go
martinsaporiti: If I return map[string][]domain.Claims I will have to iterate the claims later because the event expect for credentials ids (string). So, this cosmetic change generates an extra iteration process, not sure if we want that only for having a better or more explanatory function signature

javip97 (approved):

x1m3 (approved):

PR #346, opened 2023-04-25
fix: returning an error when trying to create qr code for invalid link

javip97:
This PR contains the validation of a link in order to generate a qr code. Invalid links will cause a 404 error

Definition of invalid link:
1. Link is not active
2. Link has reached the maximum amount 
3. Link is expired


x1m3 (approved):

martinsaporiti (approved):
LGTM!!!

PR #348, opened 2023-04-25
Refactor proofTypes domain modeling

amonsosanz:


amonsosanz:
This approach is not possible. Find [here](https://github.com/0xPolygonID/sh-id-platform/pull/342#discussion_r1176817287) more details.

PR #352, opened 2023-04-26
Issuer can view details of issued credentials

amonsosanz:
Resolves PID-598


ui/src/components/credentials/attributes/Boolean.tsx
```diff
@@ -15,6 +15,7 @@ export function Boolean({
       label={<Typography.Text>{attribute.schema.title || attribute.name}</Typography.Text>}
       name={["credentialSubject", ...parents.map((parent) => parent.name), attribute.name]}
       required={attribute.required}
```
cneuro, line 17: You don't need this flag if you use `rules`. But why do you want to do this, adding generic validation messages?

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -0,0 +1,266 @@
...
+    if (credentialID) {
+      const { aborter } = makeRequestAbortable(fetchCredential);
+      return aborter;
+    }
+    return;
+  }, [fetchCredential, credentialID]);
+
+  const loading = isAsyncTaskStarting(credential) || isAsyncTaskStarting(credentialSubjectValue);
+
+  const credentialSubjectValueErrorToString = (error: string | z.ZodError) =>
```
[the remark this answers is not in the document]
cneuro: It's a helper, right? Makes sense to me putting it in `utils/error.ts`.

ui/src/components/credentials/attributes/Boolean.tsx
cneuro: But we are letting the BE and parsers flag errors, not relying on inline form errors. If you define `rules`, it always defaults to a message.

Screenshot from 2023-04-26 19-17-27

amonsosanz: The problem with the errors from the backend is that are not targeted to the fields, then, do you want to remove all validation and let the user be aware of the errors through toast bar notifications? or are you asking for customized inline errors? I'm a bit lost.

ui/src/components/credentials/CredentialDetails.tsx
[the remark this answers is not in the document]
amonsosanz: Yep, it makes sense, at the end of the day we were already importing the generic `processZodError` util from there. This is a bit more specific but still shared anyway. Done.

adrifdez (approved):

ui/src/components/credentials/attributes/Boolean.tsx
cneuro: > let the user be aware of the required field errors through toast bar notifications

This is how we always had it. If we want to add extra form validation, then it should be everywhere (replacing the `required` flag in `Form.Item` with `rules`). The problem is that it's [buggy](https://github.com/0xPolygonID/sh-id-platform/pull/342#discussion_r1177519945), which is why I would avoid adding it right now also in the interest of time.

I recommend adding an issue to consolidate all `required`, `requiredMark` and `rules` so that form validation can be implemented properly.

amonsosanz: Done as agreed @cneuro.

cneuro: Relevant issue: https://polygonid.atlassian.net/browse/PID-651

PR #365, opened 2023-04-28
Added page for adding a directly-issued credential via QR code

adrifdez:
Resolves PID-655 

Using this requires the issuer to have direct communication with the holder that the credential was issued to, so that the link with the QR to scan the credential offer can be given to them.

### DIFF
--- a/ui/src/adapters/api/credentials.ts
+++ b/ui/src/adapters/api/credentials.ts
@@ -499,7 +499,7 @@ export async function createAuthQRCode({
   }
 }
 
-type BodyIssuedQRCode = {
+type IssuedQRCodeDescription = {
   body: {
     credentials: {
       description: string;
@@ -507,10 +507,12 @@ type BodyIssuedQRCode = {
   };
 };
 
-const bodyIssuedQRCodeParser = getStrictParser<BodyIssuedQRCode>()(
-  z.object({
-    body: z.object({ credentials: z.array(z.object({ description: z.string() })) }),
-  })
+const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeDescription, string | undefined>()(
+  z
+    .object({
+      body: z.object({ credentials: z.array(z.object({ description: z.string() })) }),
+    })
+    .transform((parsed) => parsed.body.credentials[0]?.description.split("#").pop())
 );
 
 type ResultOkIssuedQRCodeInput = {
@@ -524,18 +526,13 @@ const resultOKSIssuedQRCodeParser = getStrictParser<
 >()(
   z.object({
     data: z.unknown().transform((unknown): IssuedQRCode => {
-      const parsed = bodyIssuedQRCodeParser.safeParse(unknown);
+      const parsedSchemaType = issuedQRCodeTypeParser.safeParse(unknown);
+      const schemaType = parsedSchemaType.success ? parsedSchemaType.data : undefined;
 
-      if (parsed.success) {
-        return {
-          qrCode: unknown,
-          schemaType: parsed.data.body.credentials[0]?.description.split("#").pop(),
-        };
-      } else {
-        return {
-          qrCode: unknown,
-        };
-      }
+      return {
+        qrCode: unknown,
+        schemaType,
+      };
     }),
     status: z.literal(200),
   })
--- a/ui/src/adapters/api/credentials.ts
+++ b/ui/src/adapters/api/credentials.ts
@@ -514,7 +514,7 @@ const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeTypeInput, string | u
     .object({
       body: z.object({ credentials: z.tuple([z.object({ description: z.string() })]) }),
     })
-    .transform((parsed) => parsed.body.credentials[0].description.split("#").pop())
+    .transform((data) => data.body.credentials[0].description.split("#").pop())
 );
 
 type ResultOkIssuedQRCodeInput = {
--- a/ui/src/adapters/api/credentials.ts
+++ b/ui/src/adapters/api/credentials.ts
@@ -499,20 +499,22 @@ export async function createAuthQRCode({
   }
 }
 
-type IssuedQRCodeDescription = {
+type IssuedQRCodeTypeInput = {
   body: {
-    credentials: {
-      description: string;
-    }[];
+    credentials: [
+      {
+        description: string;
+      }
+    ];
   };
 };
 
-const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeDescription, string | undefined>()(
+const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeTypeInput, string | undefined>()(
   z
     .object({
-      body: z.object({ credentials: z.array(z.object({ description: z.string() })) }),
+      body: z.object({ credentials: z.tuple([z.object({ description: z.string() })]) }),
     })
-    .transform((parsed) => parsed.body.credentials[0]?.description.split("#").pop())
+    .transform((parsed) => parsed.body.credentials[0].description.split("#").pop())
 );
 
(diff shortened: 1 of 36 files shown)


ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -173,6 +174,11 @@ export function CredentialDetails() {
               className="centered"
               extra={
                 <Space>
+                  <Link target="_blank" to={generatePath(ROUTES.issuedQR.path, { credentialID })}>
+                    <Button icon={<IconQR />} type="link">
+                      Scan
```
cneuro, line 179: This needs to be clearer and in line with the credential link, like "View QR code"

ui/src/components/credentials/IssuedQR.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <ErrorResult error={hasFailed.message} labelRetry="Start again" onRetry={onStartAgain} />
+    );
+  }
+
+  if (!isAsyncTaskDataAvailable(issuedQRCode)) {
+    return <LoadingResult />;
+  }
+
+  return (
+    <Space align="center" direction="vertical" size="large">
```
cneuro, line 67: Let's abstract the reusable DOM between this and `CredentialLinkQR`.

ui/src/components/credentials/IssuedQR.tsx
```diff
@@ -0,0 +1,110 @@
...
+          >
+            <QRCodeSVG
+              className="full-width"
+              includeMargin
+              level="H"
+              style={{ height: 300 }}
+              value={JSON.stringify(issuedQRCode.data)}
+            />
+          </Col>
+        </Row>
```
cneuro, line 106: This must include `schemaType` i.e. name of credential.

ui/src/components/shared/Detail.tsx
```diff
@@ -18,7 +18,7 @@ export function Detail({
   text: string;
 }) {
   const value = ellipsisPosition ? text.slice(0, text.length - ellipsisPosition) : text;
-  const isUrl = z.string().url().safeParse(text).success;
+  const isUrl = text.startsWith("http://") || text.startsWith("https://");
```
cneuro, line 21: What about `ftp://`?

ui/src/routes.ts
```diff
@@ -6,6 +6,7 @@ export type RouteID =
   | "credentials"
   | "importSchema"
   | "issueCredential"
+  | "issuedQR"
```
cneuro, line 9: This is not expressive. Try something like "issuedCredentialQRCode"

ui/src/styles/index.scss
```diff
@@ -335,3 +335,7 @@ svg {
     }
   }
 }
+
+a .ant-typography {
```
cneuro, line 339: Nice catch, solving the issue for `Detail`!

ui/src/routes.ts
```diff
@@ -51,6 +52,10 @@ export const ROUTES: Routes = {
     layout: "sider",
     path: "/credentials/issue/:schemaID?",
   },
+  issuedQR: {
+    layout: "fullWidthGrey",
+    path: "/credentials/scan-issued/:credentialID",
```
cneuro, line 57: `scan-credential` would be more precise.

adrifdez: Remember to change nginx if you change this path

ui/src/adapters/api/credentials.ts
```diff
@@ -518,6 +518,46 @@ export async function createAuthQRCode({
   }
 }
 
+type ResultOKUnknown = {
+  data?: unknown;
+  status: 200;
+};
+
+const resultGetQRParser = getStrictParser<ResultOKUnknown>()(
```
amonsosanz, line 526: `resultOKUnknownParser`. What we parse (`<ResultOKUnknown>`) plus the suffix `parser`

ui/src/components/credentials/CredentialDetails.tsx
amonsosanz: I guess you mean to add the Link URL also to the details, right? Agreed.

Also, I'd name the button "QR code link" instead of "Scan"

ui/src/components/shared/Router.tsx
```diff
@@ -26,6 +26,7 @@ const COMPONENTS: Record<RouteID, ComponentType> = {
   credentials: Credentials,
   importSchema: ImportSchema,
   issueCredential: IssueCredential,
+  issuedQR: IssuedQR,
```
amonsosanz, line 29: As we have `credentialLinkQR: CredentialLinkQR` these should be `credentialIssuedQR: CredentialIssuedQR`

ui/src/routes.ts
amonsosanz: do u buy [this](https://github.com/0xPolygonID/sh-id-platform/pull/365/files#r1180662714) @cneuro?

ui/src/components/credentials/CredentialDetails.tsx
adrifdez: I've followed the old design for the Download but ok

ui/src/components/shared/Detail.tsx
amonsosanz: I think we only need to cover `http` and `https`. The list of [URL schemes](https://en.wikipedia.org/wiki/List_of_URI_schemes) is huge but why should we care about any other?

ui/src/components/credentials/CredentialDetails.tsx
amonsosanz: Fair enough. Do as you prefer.

ui/src/components/shared/Detail.tsx
adrifdez: The moment that we need more protocols, we'll add it

ui/src/components/credentials/IssuedQR.tsx
adrifdez: I don't understand what you want to reuse, let's see if this page persists in the future and we'll think about it

ui/src/routes.ts
adrifdez: I'm going to follow this pattern 
credential**Issued**QR: {
    layout: "fullWidthGrey",
    path: "/credentials/scan-**issued**/:credentialID",
  },
  credential**Link**QR: {
    layout: "fullWidthGrey",
    path: "/credentials/scan-**link**/:linkID",
  },

amonsosanz (approved):

ui/src/components/credentials/IssuedQR.tsx
amonsosanz: I'd not abstract anything here. There's little repetition and as Adri points out we don't know for how long this page will make it.

ui/src/components/credentials/IssuedQR.tsx
amonsosanz: This has been fixed.

ui/src/routes.ts
amonsosanz: Adri fixed this.

ui/src/routes.ts
amonsosanz: I initially thought the same regarding `scan-credential` but later realized that the current URL is under the path `/credentials/` so I think it's ok as it's.

 * `/credentials/scan-link/{id}`
 * `/credentials/scan-issued/{id}`

ui/src/components/credentials/IssuedQR.tsx
cneuro: Proposal for `CredentialQR` abstraction: https://github.com/0xPolygonID/sh-id-platform/pull/366

cneuro (approved):

PR #366, opened 2023-05-02
Added CredentialQR.

cneuro:


ui/src/components/credentials/CredentialQR.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <Space align="center" direction="vertical" size="large">
+      <Avatar shape="square" size={64} src={env.issuer.logo || IMAGE_PLACEHOLDER_PATH} />
+
+      <Space
+        direction="vertical"
+        style={{ padding: "0 24px", textAlign: "center", width: lg ? 800 : "100%" }}
+      >
+        <Typography.Title level={2}>
+          {env.issuer.name} wants to send you a credential
```
amonsosanz, line 33: If we're no longer using the issuer data from the API (which makes sense) then we should remove it from our type and parser and let the backend know so they also remove it.

```diff
diff --git a/ui/src/adapters/api/credentials.ts b/ui/src/adapters/api/credentials.ts
index 3c42f5d..7886fcd 100644
--- a/ui/src/adapters/api/credentials.ts
+++ b/ui/src/adapters/api/credentials.ts
@@ -446,7 +446,6 @@ type AuthQRCodeInput = Omit<AuthQRCode, "linkDetail"> & {
 };
 
 export interface AuthQRCode {
-  issuer: { displayName: string; logo: string };
   linkDetail: { proofTypes: ProofType[]; schemaType: string };
   qrCode?: unknown;
   sessionID: string;
@@ -454,10 +453,6 @@ export interface AuthQRCode {
 
 const authQRCodeParser = getStrictParser<AuthQRCodeInput, AuthQRCode>()(
   z.object({
-    issuer: z.object({
-      displayName: z.string(),
-      logo: z.string(),
-    }),
     linkDetail: z.object({ proofTypes: proofTypeParser, schemaType: z.string() }),
     qrCode: z.unknown(),
     sessionID: z.string(),
```

amonsosanz (approved):

ui/src/components/shared/SiderMenu.tsx
```diff
@@ -47,6 +47,7 @@ export function SiderMenu() {
       matchRoutes(
         [
           { path: credentialsPath },
+          { path: ROUTES.credentialDetails.path },
```
amonsosanz, line 50: Good catch!

PR #371, opened 2023-05-02
Add support for new description schema types format

amonsosanz:


ui/src/components/credentials/CredentialLinkQR.tsx
```diff
@@ -230,7 +230,7 @@ export function CredentialLinkQR() {
   return (
     <CredentialQR
       qrCode={authQRCode.data.qrCode}
-      schemaType={authQRCode.data.linkDetail.schemaType}
+      schemaTypes={[authQRCode.data.linkDetail.schemaType]}
```
cneuro, line 233: Why does this one not have several, like issued potentially does?

amonsosanz: Can't tell. This is to be asked to backend. I suspect one of them is not correct on the backend side.

ui/src/adapters/api/credentials.ts
```diff
@@ -496,20 +496,18 @@ export async function createAuthQRCode({
...
-      }
-    ];
+    credentials: {
+      description: string;
+    }[];
   };
 };
 
-const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeTypeInput, string | undefined>()(
+const issuedQRCodeTypeParser = getStrictParser<IssuedQRCodeTypeInput, string[]>()(
```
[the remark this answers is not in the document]
amonsosanz: Yep, the old code also works:

image

ui/src/components/credentials/CredentialLinkQR.tsx
cneuro: Resolved on a call. Previous validation for a `tuple` (array of one fixed element) is more correct in this case. An array of none or multiple arrays is possible for a user DID, but not for one credential ID. It should never be `undefined`.

amonsosanz: Changed.

ui/src/domain/credential.ts
```diff
@@ -19,7 +19,7 @@ export interface Credential {
 
 export interface IssuedQRCode {
   qrCode: unknown;
-  schemaType?: string;
+  schemaType: string;
```
cneuro, line 22: Since this is no longer optional, we can remove the guard on line 68:

```
        {schemaType && (
          <Row>
            <Col
              style={{
                padding: 24,
                paddingBottom: 8,
              }}
            >
              <Typography.Title ellipsis={{ tooltip: true }} level={3}>
                {schemaType}
              </Typography.Title>
            </Col>
          </Row>
        )}
```

ui/src/components/credentials/CredentialQR.tsx
```diff
@@ -10,7 +10,7 @@ export function CredentialQR({
   subTitle,
 }: {
   qrCode: unknown;
-  schemaType?: string;
+  schemaType: string;
```
cneuro, line 13: Since this is no longer optional, we can remove the guard on line 68:

```
        {schemaType && (
          <Row>
            <Col
              style={{
                padding: 24,
                paddingBottom: 8,
              }}
            >
              <Typography.Title ellipsis={{ tooltip: true }} level={3}>
                {schemaType}
              </Typography.Title>
            </Col>
          </Row>
        )}
```

amonsosanz: Right. Done.

PR #375, opened 2023-05-03
refactor: Remove Result types and model Error domain

amonsosanz:
Resolves: PID-488


ui/src/adapters/api/issuer-state.ts
```diff
@@ -111,28 +104,15 @@ export async function getTransactions({
...
-      data: {
-        failed: data.failed,
-        successful: data.successful.sort(
-          ({ publishDate: a }, { publishDate: b }) => dayjs(b).unix() - dayjs(a).unix()
-        ),
-      },
-      isSuccessful: true,
-    };
+    return buildSuccessResponse(
+      getListParser(transactionParser)
```
adrifdez, line 108: Do you think `getListParser` could do the sorting by accepting the key?

ui/src/components/credentials/CredentialLinkQR.tsx
```diff
@@ -117,7 +117,7 @@ export function CredentialLinkQR() {
     : undefined;
 
   if (hasFailed) {
-    if (hasFailed.status === 404) {
+    if (hasFailed.type === "request-error" && hasFailed.error.status === 404) {
```
adrifdez, line 120: This is not working, should be
```suggestion
   if (hasFailed.type === "request-error" && hasFailed.error.response.status === 404) {
```
image

ui/src/components/credentials/CredentialLinkQR.tsx
```diff
@@ -129,7 +129,7 @@ export function CredentialLinkQR() {
           </Typography.Text>
         </Space>
       );
-    } else if (hasFailed.status === 400) {
+    } else if (hasFailed.type === "request-error" && hasFailed.error.status === 400) {
```
adrifdez, line 132: :point_up:

ui/src/components/credentials/LinkDetails.tsx
```diff
@@ -81,29 +81,33 @@ export function LinkDetails() {
               });
             } else {
               setCredentialSubjectValue({
-                error: `The type "${parsedCredentialSubject.data.type}" is not a valid type for the attribute "credentialSubject".`,
+                error: buildAppError(
```
adrifdez, line 84: Wierd use the `buildAppError` inside the component, maybe better change the error type to string and extract the `message from the last case

cneuro (approved):
Spot check clear. Always a big fan of reducing code burden without sacrificing clarity.

ui/src/adapters/api/issuer-state.ts
amonsosanz: I thought about that and discarded the idea to avoid the extra complexity. This implies receiving an existing key of the type and also a sorting function compatible with that key, as dates are not sorted as numbers or strings. It does not worth the complexity IMO.

ui/src/components/credentials/CredentialLinkQR.tsx
amonsosanz: mmm... weird the changes in this PR should not make this fail.

amonsosanz: I see, you're right! [the previous implementation](https://github.com/0xPolygonID/sh-id-platform/pull/375/files#diff-04c057fe2e4d7a5a09df0be725644b5a8d5292db8be4d119016a0fae4f3f5067L66) was passing the content of `error.response`, not the whole error.

image

Nice catch @adrifdez! Thx! 😅

ui/src/components/credentials/CredentialLinkQR.tsx
amonsosanz: Fixed!

ui/src/components/credentials/LinkDetails.tsx
adrifdez: This `AsyncTask<ObjectAttributeValue, string>` is not linked to the refactor, this is the way that we use the error in this component, we are setting the error as we need it in some components, and we are accessing the message inside ZodError in other places. but it's fine. 
Maybe all that logic should be moved to a parser and cleaned the component.

Could you reuse it with the `CredentialDetails`? I'm seeing that is the same code. 
```
 if (parsedCredentialSubject.success) {
            if (parsedCredentialSubject.data.type === "object") {
              setCredentialSubjectValue({
                data: parsedCredentialSubject.data,
                status: "successful",
              });
            } else {
              setCredentialSubjectValue({
                error: buildAppError(
                  `The type "${parsedCredentialSubject.data.type}" is not a valid type for the attribute "credentialSubject".`
                ),
                status: "failed",
              });
            }
          } else {
            setCredentialSubjectValue({
              error: buildAppError(parsedCredentialSubject.error),
              status: "failed",
            });
          }
        } else {
          setCredentialSubjectValue({
            error: buildAppError(
              `Could not find the attribute "credentialSubject" in the object's schema.`
            ),
            status: "failed",
          });
        }
      } else {
        setCredentialSubjectValue({
          error: response.error,
          status: "failed",
        });
      }
    });
```

amonsosanz: A lot of comments...

> This `AsyncTask<ObjectAttributeValue, string>` is not linked to the refactor

Why not? This refactor applies to the error management of the whole app, this `AsyncTask<ObjectAttributeValue, string>` included.

> this is the way that we use the error in this component, we are setting the error as we need it in some components

Not sure I understand this. The way this component uses the errors has not changed. Only the shape of the error has, which now is required to be `AppError`.

> and we are accessing the message inside ZodError in other places

You're right there! I missed the usage of `processZodError` in the env context. Fixed. Now `processZodError` is not even exported.

> Maybe all that logic should be moved to a parser and cleaned the component.

Can you extend on this or provide a PoC, I'm not sure I'm following.

> Could you reuse it with the `CredentialDetails`? I'm seeing that is the same code.

Yes, credential details and link details are pretty similar but I don't want to think about the right abstraction now. Feel free to create a dedicated task so we can think about this abstraction globally, not only centered on error management.

adrifdez: To be continued https://polygonid.atlassian.net/browse/PID-682

adrifdez (approved):

PR #379, opened 2023-05-03
feat: Handling for configurable UI app-wide warning notification.

cneuro:
When deploying to test environment, needs a manual step of setting `ISSUER_UI_WARNING_MESSAGE` in `.env-ui` to the following:

> This Issuer Node is for testing purposes only. It is publicly available. Do not use personal or sensitive data. All data is deleted every 48 hours.


ui/src/components/shared/Router.tsx
```diff
@@ -47,6 +61,21 @@ export function Router() {
...
+  useEffect(() => {
+    if (env.warningMessage && isShowingWarning) {
+      notification.warning({
+        closeIcon: <IconClose />,
+        description: env.warningMessage,
+        duration: 0,
+        icon: <IconAlert />,
+        key: warningKey,
+        message: "Warning",
+        onClose: () => setShowWarning(setStorageByKey({ key: warningKey, value: false })),
```
amonsosanz, line 73: Are you sure about storing this in the LocalStorage with a hardcoded key? Since the key is not configurable at deployment time, once closed by one person on a computer the msg won't show again even if the app is redeployed and a new warning message needs to be shown.

I can think of 2 alternatives to this issue:
 * Use the SessionStorage instead, so the msg is shown once per session (tab instance)
 * Deriving the key from the content of the msg, so that whenever the content is updated, the warning appears again.

Also, have you considered a non-closable alternative?

image

cneuro: Thanks! Good point about the warning permanence between deployments, but the same could be said about having the Explainer being permanently dismissable.

The design and functionality was agreed to be directly based on what we did for PID platform deprecation (see Slack [thread](https://0xpolygon.slack.com/archives/C0409DDNV7V/p1683114137848449)). However, I am intrigued by this alternate banner mockup (non-dismissable and fixed, I presume), this seems better UX. Is this an Ant, or a custom component?

amonsosanz: I just update the comment above, sorry!

ui/src/components/shared/Router.tsx
```diff
@@ -47,6 +61,21 @@ export function Router() {
         : acc;
     }, []);
 
+  useEffect(() => {
+    if (env.warningMessage && isShowingWarning) {
```
adrifdez, line 65: I'd not show the notification on all the pages, it's not an alert for the QR code pages, I'd move it to SiderLayoutContent at this moment

amonsosanz: Yep, I was about to comment the same on this. My proposal was to add this as a prop per-route. I don't like basing this on the layout either.

cneuro: Good point!

cneuro: I think it makes sense to alert a holder that this is a test environment and more importantly that data will be deleted?

ui/src/components/shared/Router.tsx
cneuro: I've made amends based on your suggestion.

amonsosanz: awesome! I guess you don't care much about the clean-up, right?

amonsosanz: I'd also do it. Let's be polite and keep the house clean. This is a possible implementation to remove old keys when the warning is closed:

```diff
diff --git a/ui/src/components/shared/Router.tsx b/ui/src/components/shared/Router.tsx
index a2558a9..461cbf3 100644
--- a/ui/src/components/shared/Router.tsx
+++ b/ui/src/components/shared/Router.tsx
@@ -24,7 +24,12 @@ import { NotFound } from "src/components/shared/NotFound";
 import { useEnvContext } from "src/contexts/Env";
 import { Layout, ROUTES, RouteID } from "src/routes";
 import { ROOT_PATH } from "src/utils/constants";
-import { getStorageByKey, setStorageByKey } from "src/utils/localStorage";
+import {
+  getStorage,
+  getStorageByKey,
+  removeStorageByKey,
+  setStorageByKey,
+} from "src/utils/localStorage";
 
 const COMPONENTS: Record<RouteID, ComponentType> = {
   connectionDetails: ConnectionDetails,
@@ -73,7 +78,14 @@ export function Router() {
         icon: <IconAlert />,
         key: warningKey,
         message: "Warning",
-        onClose: () => setShowWarning(setStorageByKey({ key: warningKey, value: false })),
+        onClose: () => {
+          setShowWarning(setStorageByKey({ key: warningKey, value: false }));
+          Object.keys(getStorage()).forEach((key) => {
+            if (key.startsWith("warningNotification") && key !== warningKey) {
+              removeStorageByKey({ key });
+            }
+          });
+        },
         placement: "bottom",
       });
     }
diff --git a/ui/src/utils/localStorage.ts b/ui/src/utils/localStorage.ts
index f6b7c69..099e6eb 100644
--- a/ui/src/utils/localStorage.ts
+++ b/ui/src/utils/localStorage.ts
@@ -41,3 +41,11 @@ export function setStorageByKey<T>({ key, value }: { key: string; value: T }) {
 
   return value;
 }
+
+export function getStorage() {
+  return localStorage;
+}
+
+export function removeStorageByKey({ key }: { key: string }) {
+  localStorage.removeItem(key);
+}
```

ui/src/components/shared/Router.tsx
cneuro: This is re-implemented according to this [discussion](https://0xpolygon.slack.com/archives/C044J4HN360/p1683278628642289).

ui/src/components/shared/Router.tsx
cneuro: This shouldn't happen `onClose` though, rather on app-reload/page refresh. Let me tinker with this.

amonsosanz: mmmm... not sure about that. I think that it makes sense to remove old keys only when a new key is deployed.

Here it's a better and more abstract implementation of the util by the way:

```diff
diff --git a/ui/src/components/shared/Router.tsx b/ui/src/components/shared/Router.tsx
index a2558a9..1541d1d 100644
--- a/ui/src/components/shared/Router.tsx
+++ b/ui/src/components/shared/Router.tsx
@@ -24,7 +24,7 @@ import { NotFound } from "src/components/shared/NotFound";
 import { useEnvContext } from "src/contexts/Env";
 import { Layout, ROUTES, RouteID } from "src/routes";
 import { ROOT_PATH } from "src/utils/constants";
-import { getStorageByKey, setStorageByKey } from "src/utils/localStorage";
+import { getStorageByKey, removeStorageByFilter, setStorageByKey } from "src/utils/localStorage";
 
 const COMPONENTS: Record<RouteID, ComponentType> = {
   connectionDetails: ConnectionDetails,
@@ -73,7 +73,12 @@ export function Router() {
         icon: <IconAlert />,
         key: warningKey,
         message: "Warning",
-        onClose: () => setShowWarning(setStorageByKey({ key: warningKey, value: false })),
+        onClose: () => {
+          setShowWarning(setStorageByKey({ key: warningKey, value: false }));
+          removeStorageByFilter(
+            (key) => key.startsWith("warningNotification") && key !== warningKey
+          );
+        },
         placement: "bottom",
       });
     }
diff --git a/ui/src/utils/localStorage.ts b/ui/src/utils/localStorage.ts
index f6b7c69..3de966b 100644
--- a/ui/src/utils/localStorage.ts
+++ b/ui/src/utils/localStorage.ts
@@ -41,3 +41,9 @@ export function setStorageByKey<T>({ key, value }: { key: string; value: T }) {
 
   return value;
 }
+
+export function removeStorageByFilter(filter: (key: string) => boolean) {
+  Object.keys(localStorage).forEach((key) => {
+    filter(key) && localStorage.removeItem(key);
+  });
+}
```

cneuro: Improvement pushed.

ui/src/utils/constants.ts
```diff
@@ -27,6 +27,7 @@ export const SCHEMA_TYPE = "Schema type";
 export const SCHEMAS = "Schemas";
 export const STATUS = "Status";
 export const VALUE_REQUIRED = "Value required";
+export const WARNING_ID = "warningNotification";
```
amonsosanz, line 30: This does not belong to this section. It's not a UI literal. I'd at least add a line break.

amonsosanz (approved):

ui/src/components/shared/Router.tsx
amonsosanz: I'm closing this, but for the record:

I prefer this generic, cleaner, and more functional implementation, which centralizes the access to the local storage in the module localStorage and moves the iteration logic and the removal logic out of the view.

```typescript
export function removeStorageByFilter(filter: (key: string) => boolean) {
  Object.keys(localStorage).forEach((key) => {
    filter(key) && localStorage.removeItem(key);
  });
}
```

so in the view, we just need to do:

```typescript
removeStorageByFilter(
  (key) => key.startsWith("warningNotification") && key !== warningKey
);
```

It's sad I wasn't able to convince u 😞

ui/src/utils/constants.ts
adrifdez: And can be local because it's just used in one file

cneuro: The header for this section is "Literals used more than once" though. To me it makes sense to have it here in the absence of other structured ordering.

ui/src/components/shared/Router.tsx
cneuro: I get the idea of having this be reusable - for what though? And it's adding quite a few more LoC and unneeded complexity at this stage. We will revisit this in [PID-684](https://polygonid.atlassian.net/browse/PID-684).

adrifdez (approved):

ui/src/utils/constants.ts
amonsosanz: This section is intended for literals readable by the user. Maybe we can improve the comment, but a local storage key is not visible to the user in the UI.

cneuro: You are right, I've added it to the file where it's used.

PR #381, opened 2023-05-04
feat: sync selectors with url

adrifdez:
Resolves PID-654

### DIFF
--- a/ui/src/adapters/parsers/forms.ts
+++ b/ui/src/adapters/parsers/forms.ts
@@ -188,14 +188,6 @@ export const credentialFormParser = getStrictParser<
     })
 );
 
-export const linkExpirationDateParser = getStrictParser<{
-  linkExpirationDate: dayjs.Dayjs | null;
-}>()(
-  z.object({
-    linkExpirationDate: dayjsInstanceParser.nullable(),
-  })
-);
-
 export const didParser = getStrictParser<{
   did: string;
 }>()(
--- a/ui/src/adapters/parsers/forms.ts
+++ b/ui/src/adapters/parsers/forms.ts
@@ -86,7 +86,7 @@ export type IssuanceMethodFormData =
       type: "directIssue";
     };
 
-const issuanceMethodFormDataParser = getStrictParser<IssuanceMethodFormData>()(
+export const issuanceMethodFormDataParser = getStrictParser<IssuanceMethodFormData>()(
   z.union([
     linkExpirationParser.and(
       z.object({
@@ -186,14 +186,6 @@ export const credentialFormParser = getStrictParser<
         };
       }
     })
-);
-
-export const didParser = getStrictParser<{
-  did: string;
-}>()(
-  z.object({
-    did: z.string(),
-  })
 );
 
 // Serializers
--- a/ui/src/adapters/parsers/forms.ts
+++ b/ui/src/adapters/parsers/forms.ts
@@ -196,6 +196,14 @@ export const linkExpirationDateParser = getStrictParser<{
   })
 );
 
+export const didParser = getStrictParser<{
+  did: string;
+}>()(
+  z.object({
+    did: z.string(),
+  })
+);
+
 // Serializers
 
 export function serializeCredentialLinkIssuance({
--- a/ui/src/components/credentials/IssuanceMethodForm.tsx
+++ b/ui/src/components/credentials/IssuanceMethodForm.tsx
@@ -18,11 +18,7 @@ import { useCallback, useEffect, useState } from "react";
 
 import { APIError } from "src/adapters/api";
 import { getConnections } from "src/adapters/api/connections";
-import {
-  IssuanceMethodFormData,
-  didParser,
-  linkExpirationDateParser,
-} from "src/adapters/parsers/forms";
+import { IssuanceMethodFormData, linkExpirationDateParser } from "src/adapters/parsers/forms";
 import { ReactComponent as IconRight } from "src/assets/icons/arrow-narrow-right.svg";
 import { useEnvContext } from "src/contexts/Env";
 import { Connection } from "src/domain";
@@ -37,7 +33,7 @@ export function IssuanceMethodForm({
   onSubmit,
 }: {
   initialValues: IssuanceMethodFormData;
-  onChangeDid: (did: string) => void;
+  onChangeDid: (did?: string) => void;
   onSubmit: (values: IssuanceMethodFormData) => void;
 }) {
   const env = useEnvContext();
@@ -98,10 +94,10 @@ export function IssuanceMethodForm({
           ) {
             setIssuanceMethod({ ...allValues, linkExpirationTime: undefined });
           } else {
-            const parsedDid = didParser.safeParse(changedValues);
-
-            if (parsedDid.success) {
-              onChangeDid(parsedDid.data.did);
+            if (allValues.type === "directIssue") {
+              onChangeDid(allValues.did);
+            } else {
+              onChangeDid(undefined);
             }
 
             setIssuanceMethod(allValues);
--- a/ui/src/components/credentials/IssuanceMethodForm.tsx
+++ b/ui/src/components/credentials/IssuanceMethodForm.tsx
@@ -28,7 +28,6 @@ import { ACCESSIBLE_UNTIL, CREDENTIAL_LINK, VALUE_REQUIRED } from "src/utils/con
 import { processZodError } from "src/utils/error";
 
 export function IssuanceMethodForm({
-  did,
   initialValues,
   onChangeDid,
   onSubmit,
--- a/ui/src/components/credentials/IssuanceMethodForm.tsx
+++ b/ui/src/components/credentials/IssuanceMethodForm.tsx
@@ -15,7 +15,6 @@ import {
 } from "antd";
 import dayjs from "dayjs";
 import { useCallback, useEffect, useState } from "react";
-import { useSearchParams } from "react-router-dom";
 
 import { APIError } from "src/adapters/api";
 import { getConnections } from "src/adapters/api/connections";
@@ -25,32 +24,23 @@ import { useEnvContext } from "src/contexts/Env";
 import { Connection } from "src/domain";
 import { AsyncTask, isAsyncTaskDataAvailable } from "src/utils/async";
 import { makeRequestAbortable } from "src/utils/browser";
-import {
-  ACCESSIBLE_UNTIL,
-  CREDENTIAL_LINK,
-  DID_SEARCH_PARAM,
-  VALUE_REQUIRED,
-} from "src/utils/constants";
+import { ACCESSIBLE_UNTIL, CREDENTIAL_LINK, VALUE_REQUIRED } from "src/utils/constants";
 import { processZodError } from "src/utils/error";
 
 export function IssuanceMethodForm({
+  did,
   initialValues,
+  onChangeDid,
   onSubmit,
 }: {
+  did?: string;
   initialValues: IssuanceMethodFormData;
+  onChangeDid: (did: unknown) => void;
   onSubmit: (values: IssuanceMethodFormData) => void;
 }) {
   const env = useEnvContext();
-  const [searchParams] = useSearchParams();
 
-  const [issuanceMethod, setIssuanceMethod] = useState<IssuanceMethodFormData>(
-    initialValues.type === "directIssue" && !initialValues.did
-      ? {
-          ...initialValues,
-          did: searchParams.get(DID_SEARCH_PARAM) || undefined,
-        }
-      : initialValues
-  );
+  const [issuanceMethod, setIssuanceMethod] = useState<IssuanceMethodFormData>(initialValues);
   const [connections, setConnections] = useState<AsyncTask<Connection[], APIError>>({
     status: "pending",
   });
@@ -96,6 +86,7 @@ export function IssuanceMethodForm({
         onFinish={onSubmit}
         onValuesChange={(changedValues, allValues) => {
           const parsedLinkExpirationDate = linkExpirationDateParser.safeParse(changedValues);
+
           if (
             allValues.type === "credentialLink" &&
             parsedLinkExpirationDate.success &&
@@ -105,6 +96,7 @@ export function IssuanceMethodForm({
           ) {
             setIssuanceMethod({ ...allValues, linkExpirationTime: undefined });
           } else {
+            onChangeDid(changedValues);
             setIssuanceMethod(allValues);
           }
         }}
--- a/ui/src/components/credentials/IssuanceMethodForm.tsx
+++ b/ui/src/components/credentials/IssuanceMethodForm.tsx
@@ -18,7 +18,7 @@ import { useCallback, useEffect, useState } from "react";
 
 import { APIError } from "src/adapters/api";
 import { getConnections } from "src/adapters/api/connections";
-import { IssuanceMethodFormData } from "src/adapters/parsers/forms";
+import { IssuanceMethodFormData, issuanceMethodFormDataParser } from "src/adapters/parsers/forms";
 import { ReactComponent as IconRight } from "src/assets/icons/arrow-narrow-right.svg";
 import { useEnvContext } from "src/contexts/Env";
 import { Connection } from "src/domain";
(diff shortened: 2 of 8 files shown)


ui/src/components/schemas/SchemaDetails.tsx
```diff
@@ -185,7 +186,10 @@ export function SchemaDetails() {
                   <Button
                     icon={<CreditCardIcon />}
                     onClick={() => {
-                      navigate(generatePath(ROUTES.issueCredential.path, { schemaID }));
+                      navigate({
+                        pathname: generatePath(ROUTES.issueCredential.path),
+                        search: schemaID ? `?${SCHEMA_SEARCH_PARAM}=${schemaID}` : "",
```
cneuro, line 191: This should be

`search: schemaID ? new URLSearchParams(SCHEMA_SEARCH_PARAM, schemaID) : undefined`. (The `?` is never needed even when setting a literal)

If you search for "search: `" you will find other verbatim instances that should be changed to this.

ui/src/components/credentials/IssuanceMethodForm.tsx
```diff
@@ -25,32 +24,23 @@ import { useEnvContext } from "src/contexts/Env";
...
-  ACCESSIBLE_UNTIL,
-  CREDENTIAL_LINK,
-  DID_SEARCH_PARAM,
-  VALUE_REQUIRED,
-} from "src/utils/constants";
+import { ACCESSIBLE_UNTIL, CREDENTIAL_LINK, VALUE_REQUIRED } from "src/utils/constants";
 import { processZodError } from "src/utils/error";
 
 export function IssuanceMethodForm({
+  did,
```
cneuro, line 31: 'did' is defined but never used.

Is your linter working?

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -75,7 +88,47 @@ export function IssueCredential() {
   });
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
-  const { schemaID } = useParams();
+  const onChangeDid = (changedValues: unknown) => {
+    const parsedDid = didParser.safeParse(changedValues);
+
+    if (parsedDid.success) {
```
cneuro, line 94: If somehow this isn't a DID or is null, then it should be handled as well, not ignored.

ui/src/components/schemas/SchemaDetails.tsx
adrifdez: That doesn't work as you said. I'm just using `new URLSearchParams` when I want to persist other values, but that method need as well the `set` and after to`toString` like this example
```
const search = new URLSearchParams(searchParams);

      search.set(SCHEMA_SEARCH_PARAM, schema.id);

      navigate(
        {
          pathname: generatePath(ROUTES.issueCredential.path),
          search: search.toString(),
        },
        { replace: true }
      );
```

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -75,7 +88,47 @@ export function IssueCredential() {
...
+      const { did } = parsedDid.data;
+      const search = new URLSearchParams(searchParams);
+
+      if (did) {
+        search.set(DID_SEARCH_PARAM, did);
+      } else {
```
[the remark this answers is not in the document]
adrifdez: When you remove it

ui/src/components/credentials/IssueCredential.tsx
adrifdez: Why? it's called onChangeDid

ui/src/components/credentials/IssuanceMethodForm.tsx
adrifdez: Looks like the pre-commit hook is not working

ui/src/components/credentials/IssueCredential.tsx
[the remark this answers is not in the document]
cneuro: So when is it ever nor successful if the only values are a string or an empty string?

ui/src/components/credentials/IssueCredential.tsx
adrifdez: When other parameters change

ui/src/components/schemas/SchemaDetails.tsx
cneuro: Oops, you are right, I was conflating the signature of the instantiation with the `.set` method.

NB we still don't need the "?".

ui/src/components/schemas/SchemasTable.tsx
```diff
@@ -66,9 +72,10 @@ export function MySchemas() {
               Details
             </Link>
             <Link
-              to={generatePath(ROUTES.issueCredential.path, {
-                schemaID,
-              })}
+              to={{
+                pathname: generatePath(ROUTES.issueCredential.path),
+                search: `?${SCHEMA_SEARCH_PARAM}=${schemaID}`,
```
cneuro, line 77: [Ditto](https://github.com/0xPolygonID/sh-id-platform/pull/381/files#r1184865106).

ui/src/components/credentials/IssueCredential.tsx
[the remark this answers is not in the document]
adrifdez: when it's an empty string we remove it from the url

ui/src/components/schemas/SchemaDetails.tsx
adrifdez: We still need it because the `schemaID` is not checked in the render. Typescript doesn't know that if we have `schema` is because we have `schemaID`

cneuro: Also NB that this still holds:

> `SchemaDetails` should return an `ErrorResult` if `schemaID` is undefined. Then we don't need these subsequent checks.

adrifdez: It's not possible to load that page without `schemaID`

ui/src/components/credentials/IssueCredential.tsx
adrifdez: It' was there in the beginning, maybe with too much abstraction, let me know if you prefer the approach pushed

ui/src/components/credentials/IssuanceMethodForm.tsx
```diff
@@ -105,6 +95,7 @@ export function IssuanceMethodForm({
           ) {
             setIssuanceMethod({ ...allValues, linkExpirationTime: undefined });
           } else {
+            onChangeDid(changedValues);
```
amonsosanz: This callback should expect a valid `did` of type `string` and be called only after the values of the form have been parsed here (instead of having to parse in `IssueCredential.tsx` whatever value is changed in the form, as now).

adrifdez: Updated one second ago

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -196,8 +243,16 @@ export function IssueCredential() {
...
             return (
               <IssuanceMethodForm
+                did={did}
                 initialValues={credentialFormInput.issuanceMethod}
+                onChangeDid={onChangeDid}
                 onSubmit={(values) => {
+                  if (values.type === "directIssue" && values.did) {
+                    onChangeDid(values.did);
+                  } else {
+                    onChangeDid("");
```
amonsosanz, line 253: Let's better make this `undefined` or `null`

ui/src/components/credentials/IssuanceMethodForm.tsx
```diff
@@ -15,42 +15,35 @@ import {
...
-} from "src/utils/constants";
+import { ACCESSIBLE_UNTIL, CREDENTIAL_LINK, VALUE_REQUIRED } from "src/utils/constants";
 import { processZodError } from "src/utils/error";
 
 export function IssuanceMethodForm({
   initialValues,
+  onChangeDid,
   onSubmit,
 }: {
+  did?: string;
```
cneuro, line 39: [Leftover](https://github.com/0xPolygonID/sh-id-platform/pull/381#discussion_r1184840063).

ui/src/components/schemas/SchemasTable.tsx
cneuro: Check `SchemasTable`, line 77.

adrifdez: What it's happening there?

cneuro: Screenshot from 2023-05-04 14-13-16

cneuro: Nevermind, it's fixed now.

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -197,7 +244,14 @@ export function IssueCredential() {
             return (
               <IssuanceMethodForm
                 initialValues={credentialFormInput.issuanceMethod}
+                onChangeDid={onChangeDid}
                 onSubmit={(values) => {
+                  if (values.type === "directIssue" && values.did) {
```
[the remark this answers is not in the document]
amonsosanz: or better do:

`onChangeDid(values.type === "directIssue" ? values.did : undefined);`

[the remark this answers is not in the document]
amonsosanz: But if we're updating the did on every change, why is it required to do it also here? What case are we covering?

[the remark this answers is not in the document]
adrifdez: This is in the submit event, to clean the URL if they have selected the link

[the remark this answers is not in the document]
amonsosanz: mmm... but that will bring an inconsistency if you don't clean also the input, right? Do we need to clean the URL? Not sure about this.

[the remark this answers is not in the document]
adrifdez: Yes because the URL is not going to be valid if you reload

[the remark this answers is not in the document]
amonsosanz: I mean this 🔈🔛 

https://user-images.githubusercontent.com/15896177/236227114-baab66f2-3843-4286-a3b9-9d3453f75cb5.mp4

ui/src/components/schemas/SchemaDetails.tsx
```diff
@@ -136,6 +137,10 @@ export function SchemaDetails() {
     isAsyncTaskStarting(jsonSchemaTuple) ||
     isAsyncTaskStarting(contextTuple);
 
+  if (!schemaID) {
+    return <ErrorResult error="No schema id provided" />;
```
cneuro, line 141: User-facing message, I would recommend: "No schema ID provided."

ui/src/components/credentials/IssuanceMethodForm.tsx
```diff
@@ -105,6 +98,12 @@ export function IssuanceMethodForm({
           ) {
             setIssuanceMethod({ ...allValues, linkExpirationTime: undefined });
           } else {
+            const parsedDid = didParser.safeParse(changedValues);
+
+            if (parsedDid.success) {
+              onChangeDid(parsedDid.data.did);
+            }
```
amonsosanz, line 105: Maybe we can reuse the `issuanceMethodFormDataParser` we already have instead of creating a new parser just for the DID?

```typescript
const parsedIssuanceMethodFormData =
  issuanceMethodFormDataParser.safeParse(allValues);
if (
  parsedIssuanceMethodFormData.success &&
  parsedIssuanceMethodFormData.data.type === "directIssue"
) {
  onChangeDid(parsedIssuanceMethodFormData.data.did);
}
```

adrifdez: We have to do the parser to the `changedValues` to not change the URL with other changes

amonsosanz: Ok, let's at least make the parser a first-class citizen and compose it in the `IssuanceMethodFormData` as we do with `LinkExpiration` and `linkExpirationParser`:

```diff
diff --git a/ui/src/adapters/parsers/forms.ts b/ui/src/adapters/parsers/forms.ts
index 5f495d8..4365c4b 100644
--- a/ui/src/adapters/parsers/forms.ts
+++ b/ui/src/adapters/parsers/forms.ts
@@ -76,15 +76,24 @@ const linkExpirationParser = getStrictParser<LinkExpiration>()(
   })
 );
 
+interface Did {
+  did?: string;
+}
+
+export const didParser = getStrictParser<Did>()(
+  z.object({
+    did: z.string().optional(),
+  })
+);
+
 export type IssuanceMethodFormData =
   | (LinkExpiration & {
       linkMaximumIssuance?: number;
       type: "credentialLink";
     })
-  | {
-      did?: string;
+  | (Did & {
       type: "directIssue";
-    };
+    });
 
 const issuanceMethodFormDataParser = getStrictParser<IssuanceMethodFormData>()(
   z.union([
@@ -94,10 +103,11 @@ const issuanceMethodFormDataParser = getStrictParser<IssuanceMethodFormData>()(
         type: z.literal("credentialLink"),
       })
     ),
-    z.object({
-      did: z.string().optional(),
-      type: z.literal("directIssue"),
-    }),
+    didParser.and(
+      z.object({
+        type: z.literal("directIssue"),
+      })
+    ),
   ])
 );
 
@@ -196,14 +206,6 @@ export const linkExpirationDateParser = getStrictParser<{
   })
 );
 
-export const didParser = getStrictParser<{
-  did: string;
-}>()(
-  z.object({
-    did: z.string(),
-  })
-);
-
 // Serializers
 
 export function serializeCredentialLinkIssuance({
diff --git a/ui/src/components/credentials/IssuanceMethodForm.tsx b/ui/src/components/credentials/IssuanceMethodForm.tsx
index 70247a1..40eee03 100644
--- a/ui/src/components/credentials/IssuanceMethodForm.tsx
+++ b/ui/src/components/credentials/IssuanceMethodForm.tsx
@@ -37,7 +37,7 @@ export function IssuanceMethodForm({
   onSubmit,
 }: {
   initialValues: IssuanceMethodFormData;
-  onChangeDid: (did: string) => void;
+  onChangeDid: (did?: string) => void;
   onSubmit: (values: IssuanceMethodFormData) => void;
 }) {
   const env = useEnvContext();
```

adrifdez: I've pushed a change, we can remove `linkExpirationDateParser` as well

ui/src/adapters/parsers/forms.ts
```diff
@@ -188,11 +188,11 @@ export const credentialFormParser = getStrictParser<
     })
 );
 
-export const linkExpirationDateParser = getStrictParser<{
-  linkExpirationDate: dayjs.Dayjs | null;
+export const didParser = getStrictParser<{
```
amonsosanz, line 191: I guess this can be removed now

cneuro (approved):

ui/src/components/schemas/SchemaDetails.tsx
amonsosanz: Sorry to be late to the party but I prefer Adri's original approach here. As Adri mentioned, this page won't be shown if `schemaID` is not defined, so adding this JSX feels misleading:

```typescript
if (!schemaID) {
  return <ErrorResult error="No schema ID provided." />;
}
```

This will never render, but now we have an "impossible error msg" that we need to maintain and what is worst, that induces developers to think this is a legit scenario, while it's not.

I prefer something closer to the original implementation, which is, condition the navigation to the existence of the value:

```typescript
onClick={() => {
  if (schemaID) {
    navigate({
      pathname: generatePath(ROUTES.issueCredential.path),
      search: `${SCHEMA_SEARCH_PARAM}=${schemaID}`,
    });
  }
}}
```

ui/src/components/credentials/IssueCredential.tsx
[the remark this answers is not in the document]
amonsosanz: This issue has been resolved by making the DID optional in `onChangeDid`, so when there's no DID (i.e, [when the type is not credentialLink](https://github.com/0xPolygonID/sh-id-platform/pull/381/files#diff-ed504652414e1d46f1ebce0127da598e60d86ae1eca3b3b3fcaa82a9b4b0b552R99)) the param is removed from the URL.

ui/src/components/schemas/SchemaDetails.tsx
cneuro: I don't understand how this is better. In both cases, the absence of the`schemaID` is practically impossible, being a consequence of the return type of `useParams`. In the return DOM approach, if somehow this were to happen, at least we provide a human-friendly failsafe rather than corrupted data and non-functional interactions.

Is there a way to convince the compiler that could these avoid manual safeguards altogether?

amonsosanz: > if somehow this were to happen, at least we provide a human-friendly failsafe rather than corrupted data and non-functional interactions

You have a point there. Let's leave it like this.

amonsosanz (approved):

PR #401, opened 2023-05-09
feat: Implement form data validation against the JSON schema definition

amonsosanz:
Resolves PID-590

Demo:

https://github.com/0xPolygonID/issuer-node/assets/15896177/bb0ace0e-fb6d-4752-baf0-88b692428c45


ui/src/adapters/api/credentials.ts
```diff
@@ -125,7 +135,7 @@ export async function getCredentials({
 
 export type CreateCredential = {
   credentialSchema: string;
-  credentialSubject: Json;
+  credentialSubject: JsonLiteral | ObjectAttributePayload;
```
elias-garcia, line 138: Why this is param is not a `Json` anymore?

amonsosanz: I changed it to be explicit about the fact that `Json` includes arrays and we do not yet support list attributes, but it can also be kept as `Json` since `JsonLiteral | ObjectAttributePayload` is a subset of `Json`. Changed back to `Json`.

ui/src/adapters/json.ts
```diff
@@ -33,14 +33,14 @@ export async function getJsonFromUrl({
 }: {
   signal?: AbortSignal;
   url: string;
-}): Promise<Response<Json>> {
+}): Promise<Response<Record<string, unknown>>> {
```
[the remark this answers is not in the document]
amonsosanz: Similar reason as before. I changed this because this function is used to retrieve the JSON Schema and the JSON LD Context and these are not allowed to be literal values such as `"foo"`, `3`, `true`, or `null`. These JSONs can only be objects. We can leave it as `Json` though. Reverted.

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -27,24 +54,73 @@ export function IssueCredentialForm({
...
+                return addErrorToPath(acc, path, errorMsg);
+              }
+            }, {})
+          );
+        }
+      } catch (error) {
+        notifyError(buildAppError(error));
+      }
+    }
+    return false;
```
amonsosanz, line 101: There's a hell of a bug here. I haven't been able to find a solution yet.

The problem is related to antd components (InputNumber and Date specifically, but there might be more cases) returning `null` when the value is removed.

This is easy to fix today by transforming all nulls to undefined, but we need to find another solution because this does interfere with the fact that we need to support `null` attributes.

cneuro (approved):

elias-garcia (approved):

PR #402, opened 2023-05-10
chore: update Dockerfile-arm.

martinsaporiti:
### Notes
* With this change we are reducing the docker image size from 3gb to 600mg in arm64 architectures.


javip97 (approved):
LGTM

PR #403, opened 2023-05-11
feat: Responsive UI

adrifdez:


ui/src/components/layouts/SiderLayout.tsx
```diff
@@ -1,30 +1,73 @@
...
+            padding: "20px 20px",
+            position: "fixed",
+            top: 0,
+            width: "100%",
+            zIndex: 3,
+          }}
+        >
+          <Row justify="space-between">
+            <LogoLink />
+            <IconMenu onClick={() => setCollapsed(!collapsed)} />
```
elias-garcia, line 32: The menu icon is not properly centered. This is due to the padding set in the Header, which is 20px. The height of the header is 64px, and if you subtract 40px (padding top + bottom), the space left for the icon is 24px, but the real height of the icon is 28px, so it runs out of space and makes it not being centered properly:

Captura de pantalla 2023-05-11 a las 19 55 41

Apart from that I'd wrap the svg into a `button` so it has `cursor: pointer` CSS prop to match the rest of the buttons in the app.

ui/src/components/schemas/Schemas.tsx
```diff
@@ -16,23 +16,27 @@ export function Schemas() {
...
       extra={
-        <Space align="start" size="middle">
-          <Button
-            icon={<IconCreditCardPlus />}
-            onClick={() => navigate(generatePath(ROUTES.issueCredential.path))}
-            type="default"
-          >
-            {ISSUE_CREDENTIAL}
-          </Button>
+        <Row gutter={[8, 8]}>
```
elias-garcia, line 19: Maybe we can add a little bit of space between the paragraph and the buttons?

Captura de pantalla 2023-05-11 a las 20 00 54

ui/src/components/schemas/Schemas.tsx
```diff
@@ -47,7 +51,7 @@ export function Schemas() {
           title="Credential schemas explained"
         />
 
-        <MySchemas />
+        <SchemasTable />
```
elias-garcia, line 54: Currently, the `paddingLeft` and the `paddingRight` is different at each level of the application. The side padding in the header is different than the side padding of the title and the description on the page, which at the same time is different than the padding of the table. Ee should use the same padding for all the items in the screen to make it beautiful for the sight (all should be alined to the red line in this screenshot, for example).

Captura de pantalla 2023-05-11 a las 19 59 25

ui/src/components/layouts/SiderLayout.tsx
adrifdez: I wrapped it into a button but it gives more padding and lose more control over the position. 
I'm going to try a pixel-perfect version

ui/src/components/schemas/Schemas.tsx
adrifdez: I'm going to respect the design that has more padding 
image

ui/src/components/schemas/Schemas.tsx
adrifdez: Yes, the problem is not there, but I'll add it

ui/src/components/schemas/Schemas.tsx
elias-garcia: I've just seen that in the designs that red line is not respected neither... 😶‍🌫️

adrifdez: I've pushed a change to align everything, let me know your opinion

ui/src/components/shared/SiderLayoutContent.tsx
```diff
@@ -1,4 +1,4 @@
-import { Button, Col, Divider, Layout, Row, Space, Typography, notification } from "antd";
+import { Button, Col, Divider, Grid, Layout, Row, Typography, notification } from "antd";
```
amonsosanz, line 1: I see this in the console

image

ui/src/components/shared/SiderLayoutContent.tsx
```diff
@@ -62,36 +64,35 @@ export function SiderLayoutContent({
...
   return (
-    <>
-      <Layout.Header
-        className="bg-light"
-        style={{ height: "auto", padding: 32, paddingBottom: showDivider ? 0 : 12 }}
+    <Layout className="bg-light" style={{ minHeight: "100vh" }}>
+      <Row
+        gutter={[0, 16]}
+        justify="space-between"
+        style={{ height: "auto", padding: md ? 32 : 16, paddingBottom: showDivider ? 0 : 12 }}
```
amonsosanz, line 71: I see this in the console

image

ui/src/components/credentials/LinksTable.tsx
```diff
@@ -86,7 +89,7 @@ export function LinksTable() {
       ),
       sorter: ({ active: a }, { active: b }) => (a === b ? 0 : a ? 1 : -1),
       title: "Active",
-      width: 100,
+      width: md ? 100 : 60,
```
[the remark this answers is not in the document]
adrifdez: The default width is a percentage between numbers of columns but It'll give a lot of white space in some cases and resizing the screen is not a normal user behavior

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -1,4 +1,4 @@
-import { Button, Card, Space, Typography } from "antd";
+import { Button, Card, Col, Grid, Row, Space, Typography } from "antd";
```
amonsosanz, line 1: The page credential details has some room for improvement

image

  * The buttons Revoke and Delete could probably be inline on this screen, but if the stack, I'm missing some margin.
  * The Credential Details box could look better. We have a similar issue with this box in Schema Details. See below:

image

amonsosanz: This is now working better but I see that something breaks on resize:

https://github.com/0xPolygonID/issuer-node/assets/15896177/0a43793c-dbf4-45c6-a8c8-20b698b6e493

amonsosanz: Per my tests, commenting [this line](https://github.com/0xPolygonID/issuer-node/blob/PID-14-responsive-ui/ui/src/components/shared/Detail.tsx#L30) fixes the issue.

amonsosanz: Regarding the text not recovering its original size, [there's a bug opened in antd](https://github.com/ant-design/ant-design/issues/32899).

amonsosanz: Yep, it looks definitely related. If I remove the copyable prop from the `Typography.Text`, I get no exceptions and no resize problems. We should consider implementing our own copy button. What a crappy lib.

ui/src/components/layouts/SiderLayout.tsx
elias-garcia: Now it looks perfect!

ui/src/components/credentials/CredentialDetails.tsx
adrifdez: Uff since v4, so what is your idea? Follow in a new issue?

amonsosanz: I'd stop using the copy feature of antd and implement our own.

ui/src/styles/index.scss
```diff
@@ -178,8 +185,23 @@
   margin-right: 16px;
 }
 
-.ant-layout .ant-layout-header {
-  line-height: inherit;
+.ant-layout {
+  .responsive-header {
+    line-height: inherit;
+    border-bottom: 1px solid #eaecf0;
```
cneuro, line 191: This should be defined and used as a new variable, e.g. `$divider-color: #eaecf0;`.

Also noticing another hardcoded color value on line 168 - it should instead be `color: $error-color;`.

ui/src/components/credentials/CredentialDetails.tsx
amonsosanz: You're on holiday man... don't you prefer to hand off this PR? I can take it from here if you agree.

Maybe we can also draft it until you're back. I don't think this is urgent. What do you think @cneuro?

cneuro (approved):

ui/src/components/credentials/CredentialDetails.tsx
cneuro: It's not a new bug, we can fix it later properly. issue: https://polygonid.atlassian.net/browse/PID-735?atlOrigin=eyJpIjoiNGRiOTYzNjkzYmJjNGIzNDgwNGU4NWJlNWNmYjY0ZDIiLCJwIjoiaiJ9

amonsosanz: @cneuro [has decided](https://0xpolygon.slack.com/archives/C044J4HN360/p1684253238191329?thread_ts=1684251872.147749&cid=C044J4HN360) to create [a new issue](https://polygonid.atlassian.net/browse/PID-735) to fix this error.

amonsosanz (approved):

PR #422, opened 2023-06-16
chore: Remove duplicated notification subtest: 'should get an error,wrong credential id'

stefanlatinovic:
Noticed this duplicated subtest while exploring the codebase.


x1m3 (approved):
Thanks for the suggestion @stefanlatinovic

PR #426, opened 2023-06-22
Load token vault and did from k8s mounted volume

x1m3:
If vault or did env vars are not set, we will try to download from the mounted volume we use in k8s configuration. If not found we will fail.


x1m3:
LGTM

martinsaporiti (approved):
LGTM

PR #435, opened 2023-06-29
feat: Add support for advanced props

amonsosanz:
This PR adds support for the `default` and `const` advanced keywords of JSON Schema so that when the schema specifies a `default` value for an attribute, this value appears prepopulated in the form when issuing the credential. When the schema specifies a `const` value, the input is also disabled so the user cannot provide a different value.


ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -213,182 +215,266 @@ export function IssueCredentialForm({
...
+            if (value === undefined) {
+              return acc;
+            }
+            switch (curr.schema.format) {
+              case "date":
+              case "date-time": {
+                return { ...acc, [curr.name]: dayjs(value) };
+              }
+              case "time": {
+                return { ...acc, [curr.name]: dayjs(`1978-01-29T${value}`) };
```
elias-garcia, line 257: What is this `dayjs(`1978-01-29T${value}`)`? 🤣

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -213,182 +215,266 @@ export function IssueCredentialForm({
...
-        layout="vertical"
-        onFinish={(values: IssueCredentialFormData) => {
-          const jsonSchemaData = isAsyncTaskDataAvailable(jsonSchema) ? jsonSchema.data : undefined;
-          const credentialSubjectAttributeWithoutId =
-            jsonSchemaData && extractCredentialSubjectAttributeWithoutId(jsonSchemaData);
-
-          if (
-            jsonSchemaData &&
-            credentialSubjectAttributeWithoutId &&
+      {initialValuesWithSchemaValues ? (
```
elias-garcia, line 304: I see one problem here. If the `credentialSubjectAttributeWithoutId` is `undefined` because the `jsonSchema` failed to load, it will make this `initialValuesWithSchemaValues` to be `false`, so the `<LoadingResult />` will be shown forever.

I think that instead, this `initialValuesWithSchemaValues` variable should be an `AsyncTask` to handle the errors properly in the view.

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -213,182 +215,266 @@ export function IssueCredentialForm({
     [env, fetchJsonSchema, initialValues.schemaID, messageAPI]
   );
 
+  const computeFormObjectInitialValues = (
+    objectAttribute: ObjectAttribute,
```
elias-garcia, line 219: Shouldn't this be `objectAttribute?: ObjectAttribute`?

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -213,182 +215,266 @@ export function IssueCredentialForm({
...
+    initialValues: Record<string, unknown>
+  ): Record<string, unknown> | undefined => {
+    return objectAttribute.schema.attributes?.reduce(
+      (acc: Record<string, unknown>, curr: Attribute): Record<string, unknown> => {
+        switch (curr.type) {
+          case "boolean": {
+            const parsedConst = z.boolean().safeParse(curr.schema.const);
```
elias-garcia, line 226: Why are we parsing the `const` and `default` values as `unknown` in the `JsonSchema` parser and then we are parsing them again here to `boolean`, `number` or `string`? I think that we should already parse them correctly in the `JsonSchema` parser.

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: My birthday

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: Very good catch! This is indeed an ugly bug that breaks the app when there's no schema id in the URL, since the schema selector is not shown when this is the case.

amonsosanz: Fixed. Loader removed. Now we set the values to the form with a `useEffect` when they are available.

Thx for catching this one! 😅

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: mmm.. why?

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: The [JSON Schema standard allows the constant and default values to have any shape](http://json-schema.org/understanding-json-schema/reference/generic.html#annotations), including shapes that do not validate against the schema. This means that valid schemas can include non-valid values in these keywords. Applying this proposal would make valid schemas fail zod validation.

ui/src/components/credentials/IssueCredentialForm.tsx
elias-garcia: Because when you call this in the same file, `credentialSubjectAttributeWithoutId` can be `undefined`:

```ts
const credentialSubjectAttributeWithoutId = isAsyncTaskDataAvailable(jsonSchema)
  ? extractCredentialSubjectAttributeWithoutId(jsonSchema.data)
  : undefined;

useEffect(() => {
  const initialValuesWithSchemaValues: Store = credentialSubjectAttributeWithoutId
    ? {
        ...initialValues,
        credentialSubject: computeFormObjectInitialValues(
          credentialSubjectAttributeWithoutId,
          initialValues.credentialSubject || {}
        ),
      }
    : initialValues;
  form.setFieldsValue(initialValuesWithSchemaValues);
}, [computeFormObjectInitialValues, credentialSubjectAttributeWithoutId, form, initialValues]);
```

ui/src/components/credentials/IssueCredentialForm.tsx
elias-garcia: 🤣 but i mean, what's the purpose of this random date to concatenate the time?

ui/src/components/credentials/IssueCredentialForm.tsx
elias-garcia: Mmmm maybe I explained myself wrong. What I wanted to say is to move this code to a zod `transform` when parsing the `JsonSchema`. It shouldn't fail, but we can add a `defaultValue` to the `Attribute` type and just set it there.

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: Nope, it can't. When `credentialSubjectAttributeWithoutId` is `undefined` we do not call `computeFormObjectInitialValues`. Doing so would raise a compile error!

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: Let's discuss this at a meeting because I don't see how this can work.

elias-garcia (approved):

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: Changed to `1970-01-01` as agreed.

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: We agreed to create another issue to consider these changes.

PR #437, opened 2023-06-30
fix: IPFS schemas not accepted when creating links

x1m3:
The credential validator when creating links was failing with IPFS schemas


javip97 (approved):
LGTM

PR #460, opened 2023-07-05
feature: Add support for the schema version metadata

amonsosanz:


ui/src/adapters/api/schemas.ts
```diff
@@ -16,28 +16,40 @@ const apiSchemaParser = getStrictParser<ApiSchemaInput, ApiSchema>()(
...
   env: Env;
   jsonLdType: JsonLdType;
   schemaUrl: string;
+  title?: string;
+  version?: string;
 }): Promise<Response<ID>> {
   try {
     const response = await axios({
       baseURL: env.api.url,
       data: {
```
elias-garcia, line 47: I think that this could be just:

```ts
      data: {
        description: description || null,
        schemaType: jsonLdType.name,
        title: title || null,
        url: schemaUrl,
        version: version || null,
      },
```

elias-garcia (approved):

ui/src/adapters/api/schemas.ts
amonsosanz: Wouldn't that cause legit empty strings to become `null`?

image

elias-garcia: But this `title`, `description` and `version` fields are required right? We shouldn't send empty strings for them right?

amonsosanz: Nope none of them are required.

PR #461, opened 2023-07-07
fix: How to build locally typo

x1m3:
The step to install the issuer node in local mode setup was wrong. The command to compile the binaries locally is 
`make build-local` and not make build.


cerberushades (approved):

martinsaporiti (approved):
LGTM

PR #462, opened 2023-07-11
feat: making title and description nullables in schemas

javip97:
as required, we made optional the title and description fields in the schema import so as it is in the get schema and get schemas endpoints


x1m3 (approved):

martinsaporiti (approved):
LGTM

PR #474, opened 2023-08-07
Delete volumes

martinsaporiti:


Makefile
```diff
@@ -205,3 +200,21 @@ restart-ui: rm-issuer-imgs up run run-ui
 
 .PHONY: restart-ui-arm
 restart-ui-arm: rm-issuer-imgs up run-arm run-ui-arm
+
+
+# usage: make new_password=xxx change-this-default-password
+.PHONY: change-this-default-password
+change-this-default-password:
```
x1m3, line 207: Not sure if a user will understand what password is changing here. Maybe "change-vault-password" could make easier to understand?

cmd/issuer_initializer/main.go
```diff
@@ -43,9 +38,28 @@ func main() {
...
+			log.Error(ctx, "cannot init vault client with Kubernetes Auth: ", "err", err)
+			return
+		}
+	} else {
+		vaultCli, err = providers.NewVaultClient(cfg.KeyStore.Address, cfg.KeyStore.Token)
+		if err != nil {
+			log.Error(ctx, "cannot init vault client: ", "err", err)
+			return
+		}
+	}
```
x1m3, line 54: This code seems to be repeated in some places in the PR. I think we could create an internal package with a constructor that accepts the config and returns a vault client. Something similar to redis.Open() or kms.Open()

cmd/notifications/main.go
```diff
@@ -63,8 +62,31 @@ func main() {
...
+			return
+		}
+	} else {
+		log.Info(ctx, "Vault userpass auth not enabled")
+		vaultCli, err = providers.NewVaultClient(cfg.KeyStore.Address, cfg.KeyStore.Token)
+		if err != nil {
+			log.Error(ctx, "cannot init vault client: ", "err", err)
+			return
+		}
+	}
```
x1m3, line 80: Same comment as in cmd/issuer_initializer/main.go.   I think we can reuse this code

cmd/notifications/main.go
```diff
@@ -134,3 +151,23 @@ func newCredentialsService(cfg *config.Configuration, storage *db.Storage, cache
...
+		log.Info(ctx, "Issuer Did from vault", "did", cfg.APIUI.Issuer)
+		issuerDID, err := core.ParseDID(cfg.APIUI.Issuer)
+		if err != nil {
+			log.Error(ctx, "invalid issuer did format", "error", err)
+			return err
+		}
+		cfg.APIUI.IssuerDID = *issuerDID
+	}
+	return nil
+}
```
x1m3, line 173: I see this function repeated i some places. I think we could move it to a common package and reuse it. IMHO, a good place could be in config package, because it is checking an updating the configuration

cmd/pending_publisher/main.go
```diff
@@ -78,10 +75,21 @@ func main() {
...
+			log.Error(ctx, "cannot init vault client with Kubernetes Auth: ", "err", err)
+			return
+		}
+	} else {
+		log.Info(ctx, "Vault userpass auth is not enabled")
+		vaultCli, err = providers.NewVaultClient(cfg.KeyStore.Address, cfg.KeyStore.Token)
+		if err != nil {
+			log.Error(ctx, "cannot init vault client: ", "err", err)
+			return
+		}
```
x1m3, line 92: Same comment as in cmd/issuer_initializer/main.go. I think we can reuse this code

cmd/platform/main.go
```diff
@@ -71,10 +72,19 @@ func main() {
...
+		if err != nil {
+			log.Error(ctx, "cannot init vault client with Kubernetes Auth: ", "err", err)
+			return
+		}
+	} else {
+		vaultCli, err = providers.NewVaultClient(cfg.KeyStore.Address, cfg.KeyStore.Token)
+		if err != nil {
+			log.Error(ctx, "cannot init vault client: ", "err", err)
+			return
+		}
```
x1m3, line 87: Same comment as in cmd/issuer_initializer/main.go. I think we can reuse this code

cmd/platform_ui/main.go
```diff
@@ -79,10 +81,19 @@ func main() {
...
+		if err != nil {
+			log.Error(ctx, "cannot init vault client with Kubernetes Auth: ", "err", err)
+			return
+		}
+	} else {
+		vaultCli, err = providers.NewVaultClient(cfg.KeyStore.Address, cfg.KeyStore.Token)
+		if err != nil {
+			log.Error(ctx, "cannot init vault client: ", "err", err)
+			return
+		}
```
x1m3, line 96: Same comment as in cmd/issuer_initializer/main.go. I think we can reuse this code

internal/config/config.go
```diff
@@ -185,32 +179,44 @@ func (c *Configuration) SanitizeAPIUI(ctx context.Context) (err error) {
...
+}
 
-	issuerDID, err := core.ParseDID(c.APIUI.Issuer)
-	if err != nil {
-		log.Error(ctx, "invalid issuer did format", "error", err)
-		return fmt.Errorf("invalid issuer did format")
+// CheckDID checks if the issuer did is provided in the configuration file. If not, it tries to get it from vault.
+func CheckDID(ctx context.Context, cfg Configuration, vaultCli *api.Client) error {
+	log.Info(ctx, "Checking issuer did value", "did", cfg.APIUI.Issuer)
+	if cfg.APIUI.Issuer == "" {
```
javip97, line 204: As the IssuerDID can be provided in the config file I would not delete it from the sample, The rest of the PR LGTM, awesome job!

https://github.com/0xPolygonID/issuer-node/pull/474/files#diff-b9c16d00e36e5e4426865988b887442b355f46be62264002d2116ea5bf21f352L7

javip97 (approved):

javip97 (approved):

PR #496, opened 2023-09-28
Pid 1368 issuer node update zk prover and verifier

martinsaporiti:


internal/core/services/revocation.go
```diff
@@ -47,30 +58,53 @@ func NewRevocationService(ethStore StateStore, contract common.Address) *Revocat
...
-			return nil, protocol.ErrStateNotFound
-		}
+func (r *Revocation) Status(ctx context.Context, credStatus interface{}, userDID, issuerDID *w3c.DID, issuerData *verifiable.IssuerData) (*verifiable.RevocationStatus, error) {
+	status, err := convertCredentialStatus(credStatus)
+	if err != nil {
+		return nil, err
+	}
+
+	switch status.Type {
+	case verifiable.Iden3ReverseSparseMerkleTreeProof:
```
x1m3, line 68: I think that case *verifiable.RHSCredentialStatus has been lost?

x1m3:
@martinsaporiti I guess we should close this PR.  It is really outdated and with tons of conficts.

PR #503, opened 2023-10-09
chore: add w3c loader

martinsaporiti:
* Implement a w3c document loader to avoid fetching documents from an external source


javip97 (approved):

x1m3 (approved):
Very good work!

My only concern is that we will need to do some small changes to merge into devel

vmidyllic (approved):

vmidyllic:
LGTM, but not sure why do you init ipfs client with gateway. But if it's meant to be - ok.

PR #507, opened 2023-10-29
The postgres container does not work the current port and also IPFS url missing from .env-ui

Abhishek-2416:


amonsosanz:
Thx for the PR!

I also added the missing IPFS URL to the env [here](https://github.com/0xPolygonID/issuer-node/pull/495/files)

PR #552, opened 2023-11-21
feat: Backward support for old raw QRCodes in API UI

x1m3:
This PR enables backward compatibility with raw QR Codes. 

v2.3.0 introduced  the new QR codes with redirection. Instead of returning the raw QR code, this QRs include only a link to fetch real data.

With this PR, caller can pass a query param to ask for a legacy QR code. 

This can be helpful for wallets that still need to migrate to this new QR codes.


api_ui/api.yaml
```diff
@@ -1027,7 +1060,7 @@ components:
         - qrCodeLink
         - schemaType
       properties:
-        qrCodeLink:
+        qrCodeLink: # TODO: Rename to QRCode. It can be a Link or a Raw QR
```
martinsaporiti, line 1063: this TODO here is correct?

martinsaporiti (approved):
LGTM

api_ui/api.yaml
x1m3: I will remove. Ideally we should change it because we always return a QR code, no matter if it is a link or a raw qr. But renaming it would be a breaking change for ui.

PR #553, opened 2023-11-21
feat: adding authentication session

javip97:
- [x] Create new table authentication_sessions with the connection_id and session_id
- [x] Create new endpoint that returns the connection of a given session_id
- [x] Change GetAuthQRCode response

image

### DIFF
--- a/api_ui/api.yaml
+++ b/api_ui/api.yaml
@@ -41,6 +41,34 @@ paths:
           $ref: '#/components/responses/500'
 
   #authentication
+  /v1/authentication/sessions/{id}:
+    get:
+      summary: Get Authentication Connection
+      operationId: getAuthenticationConnection
+      description: get authentication connection
+      parameters:
+        - $ref: '#/components/parameters/id'
+      tags:
+          - Auth
+          - Connection
+      security:
+        - basicAuth: [ ]
+      responses:
+        '200':
+          description: ok
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/GetAuthenticationConnectionResponse'
+        '400':
+          $ref: '#/components/responses/400'
+        '401':
+          $ref: '#/components/responses/401'
+        '404':
+          $ref: '#/components/responses/404'
+        '500':
+          $ref: '#/components/responses/500'
+
   /v1/authentication/qrcode:
     get:
       summary: Get Connection QRCode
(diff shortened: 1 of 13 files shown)


api_ui/api.yaml
```diff
@@ -1084,6 +1121,11 @@ components:
           x-omitempty: false
           example: c79c9c04-8c98-40f2-a7a0-5eeabf08d836
 
+    UUIDString:
+      type: string
+      x-omitempty: false
+      example: did:polygonid:polygon:mumbai:2qFpPHotk6oyaX1fcrpQFT4BMnmg8YszUwxYtaoGoe
```
x1m3, line 1127: Example is wrong? This is not an UUID

api_ui/api.yaml
```diff
@@ -1302,6 +1344,34 @@ components:
...
+      type: object
+      required:
+        - id
+        - userID
+        - issuerID
+        - createdAt
+        - modifiedAt
+      properties:
+        id:
+          $ref: '#/components/schemas/UUIDString'
```
x1m3, line 1357: Don't you think it is better to return it as uuid instead of string as we do in other places. Like in Link
          type: string
          x-go-type: uuid.UUID
          x-go-type-import:
            name: uuid
            path: github.com/google/uuid
          example: 8edd8112-c415-11ed-b036-debe37e1cbd6

internal/core/ports/identity_service.go
```diff
@@ -39,7 +39,7 @@ type IdentityService interface {
 	UpdateIdentityState(ctx context.Context, state *domain.IdentityState) error
 	GetTransactedStates(ctx context.Context) ([]domain.IdentityState, error)
 	GetStates(ctx context.Context, issuerDID w3c.DID) ([]domain.IdentityState, error)
-	CreateAuthenticationQRCode(ctx context.Context, serverURL string, issuerDID w3c.DID) (string, error)
+	CreateAuthenticationQRCode(ctx context.Context, serverURL string, issuerDID w3c.DID) (string, string, error)
```
x1m3, line 42:
```suggestion
	CreateAuthenticationQRCode(ctx context.Context, serverURL string, issuerDID w3c.DID) (string, uuid.UUID, error)
```
Returning uuid gives more info to the caller.

x1m3 (commented):
LGTM, But it would narrow the type of some vars and return types from string to uuid to give more readability

internal/core/ports/identity_service.go
javip97: sure, done!

x1m3 (approved):

martinsaporiti (approved):
LGTM

PR #564, opened 2023-11-28
feat: Add pagination to the credentials table

amonsosanz:
Resolves PID-1573

### DIFF
--- a/ui/src/adapters/api/credentials.ts
+++ b/ui/src/adapters/api/credentials.ts
@@ -3,10 +3,15 @@ import { z } from "zod";
 
 import { Response, buildErrorResponse, buildSuccessResponse } from "src/adapters";
 import { ID, IDParser, Message, buildAuthorizationHeader, messageParser } from "src/adapters/api";
-import { datetimeParser, getListParser, getStrictParser } from "src/adapters/parsers";
+import {
+  datetimeParser,
+  getListParser,
+  getResourceParser,
+  getStrictParser,
+} from "src/adapters/parsers";
 import { Credential, Env, IssuedQRCode, Json, Link, LinkStatus, ProofType } from "src/domain";
 import { API_VERSION, QUERY_SEARCH_PARAM, STATUS_SEARCH_PARAM } from "src/utils/constants";
-import { List } from "src/utils/types";
+import { List, Resource } from "src/utils/types";
 
 type ProofTypeInput = "BJJSignature2021" | "SparseMerkleTreeProof";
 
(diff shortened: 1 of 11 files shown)


mmilenkovic (approved):

ui/src/utils/types.ts
```diff
@@ -1,8 +1,19 @@
 import { z } from "zod";
 
+export type Meta = {
```
elias-garcia, line 3: I would call this `ResourceMeta`, `Meta` sounds very generic.

ui/src/adapters/api/credentials.ts
```diff
@@ -86,17 +91,19 @@ export async function getCredential({
 
 export async function getCredentials({
   env,
-  params: { did, query, status },
+  params: { did, max_results, page, query, status },
```
elias-garcia, line 94: I'd avoid having snake case params in our interfaces. If we need to send it to the API I'd name this param `maxResults` and later send the `max_results` name to the API.

ui/src/adapters/api/credentials.ts
```diff
@@ -108,15 +115,20 @@ export async function getCredentials({
...
     return buildSuccessResponse(
-      getListParser(credentialParser)
-        .transform(({ failed, successful }) => ({
-          failed,
-          successful: successful.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime()),
+      getResourceParser(credentialParser)
+        .transform(({ items: { failed, successful }, meta }) => ({
+          items: {
+            failed,
+            successful: successful.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime()),
```
elias-garcia, line 129: I'd remove this sorting by the moment as we agreed.

ui/src/adapters/api/credentials.ts
```diff
@@ -86,17 +91,19 @@ export async function getCredential({
...
   env,
-  params: { did, query, status },
+  params: { did, max_results, page, query, status },
   signal,
 }: {
   env: Env;
   params: {
     did?: string;
+    max_results?: number;
+    page?: number;
```
elias-garcia, line 101: Why `maxResults` and `page` are optional?

ui/src/adapters/api/credentials.ts
amonsosanz: Yep, changed!

ui/src/adapters/api/credentials.ts
amonsosanz: Done

ui/src/adapters/api/credentials.ts
amonsosanz: Why should they be required? The API applies some default values.

elias-garcia (approved):

PR #575, opened 2023-12-05
feat: Pagination in get credentials api endpoint

x1m3:


javip97 (approved):
Nice Job!!

amonsosanz:
There's an issue with the pagination, at least when the DB is empty.

Initially I do this request `credentials?max_results=10&page=1` and the response I get has this meta `{"items":[],"meta":{"max_results":0,"page":1,"total":0}}`

Since we store the meta from the response in the URL, the next request we do is `credentials?max_results=0&page=1`, which is wrong and produces the error `{message: "maxResults param must be higher than 0"}`

PR #580, opened 2023-12-14
feat: adding paging to the getConnections endpoint

javip97:
this PR enables pagination in the getConnections endpoint, same implementation as in the getCredentials


martinsaporiti (approved):
LGTM

PR #581, opened 2023-12-14
feat: support refresh service

ilya-korotya:
We have recently implemented a refresh service for short-lived credentials. If a credential includes a **refreshService** field, clients can now use this service to refresh credentials that have expired.

**Typical Use Case:**

1. A user obtains a credential for their Ethereum (ETH) balance. We do not issue long-term balance credentials due to the rapid fluctuation in balances.
2. The issued balance credential expires after 5 minutes. Before expiration, the user can utilize this credential. 
3. After expiration, instead of requesting a new balance credential from the issuer, the user can simply refresh the existing credential using the refresh service.

For more details, please refer to our internal documentation: [Refresh Service Documentation](https://www.notion.so/polygontechnology/Refresh-service-b6171025f26f48ce8c3127aa0b79f40e).

~~Link to the refresh service (⚠️  integration with the issuer-node is pending): [GitHub - Refresh Service](https://github.com/0xPolygonID/refresh-service).~~
Integration with the issuer-node is ready on refresh service side: https://github.com/0xPolygonID/refresh-service/pull/3


internal/core/services/claims.go
```diff
@@ -625,9 +626,34 @@ func (c *claim) createVC(ctx context.Context, claimReq *ports.CreateClaimRequest
...
+	guards := []guardFunc{
+		// check if schema's URL is valid
+		func() error {
+			if _, err := url.ParseRequestURI(req.Schema); err != nil {
+				return ErrMalformedURL
+			}
+			return nil
+		},
+		// check if refresh service has supported type
+		func() error {
```
javip97, line 640: If RefreshService is set, expiration should be mandatory in that case, right? I'm missing that validation here

Taken from the docs: "Also, the issuer should set expiration field"

internal/db/schema/migrations/20231214192507_add_refresh_service_to_link_table.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
```
javip97, line 1: I'm missing the same migration for the `claims` table + the repository that saves the refresh service

internal/core/domain/link.go
```diff
@@ -74,6 +75,7 @@ type Link struct {
 	Active                   bool
 	Schema                   *Schema
 	IssuedClaims             int // TODO: Give a value when link redemption is implemented
+	RefreshService           *verifiable.RefreshService
```
javip97, line 78: missing this field in `domain.Claim` struct as well. This field should be populated at least in the repository function `func (c *claims) GetByIdAndIssuer` so mobile can fetch this credential with the proper RefreshService field

javip97 (commented):
Nice Job! Some minor comments

internal/core/domain/link.go
martinsaporiti: +1

javip97 (approved):
LGTM

internal/core/services/claims.go
ilya-korotya: Yes, you are right. I will fix.

PR #589, opened 2023-12-29
chore: Update go version to 1.21 and use standard slog

x1m3:
Updating go version to 1.21 and use new standard slog library instead of the experimental one.

Updating some packages to newest version


martinsaporiti (approved):
LGTM

martinsaporiti (approved):
LGTM

javip97 (approved):
LGTM

PR #594, opened 2024-01-09
feat: adding revoke push notification

javip97:
waiting for mobile implementation

16-01 -> Already tested with mobile and working

### DIFF
--- a/internal/core/ports/claims_service.go
+++ b/internal/core/ports/claims_service.go
@@ -207,6 +207,7 @@ type GetCredentialQrCodeResponse struct {
 // ClaimsService is the interface implemented by the claim service
 type ClaimsService interface {
 	Save(ctx context.Context, claimReq *CreateClaimRequest) (*domain.Claim, error)
+	GetRevoked(ctx context.Context, currentState string) ([]*domain.Claim, error)
 	CreateCredential(ctx context.Context, req *CreateClaimRequest) (*domain.Claim, error)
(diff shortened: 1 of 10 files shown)


internal/core/ports/claims_service.go
```diff
@@ -207,6 +207,7 @@ type GetCredentialQrCodeResponse struct {
 // ClaimsService is the interface implemented by the claim service
 type ClaimsService interface {
 	Save(ctx context.Context, claimReq *CreateClaimRequest) (*domain.Claim, error)
+	GetRevoked(ctx context.Context, currentState string) ([]*domain.Claim, error)
```
x1m3, line 210:
```suggestion
	GetRevoked(ctx context.Context, currentState string) ([]domain.Claim, error)
```
I find it more clear if we return []domain.Claim instead of slice of pointers, as this list should never contain nil values

martinsaporiti (approved):
LGTM

x1m3 (changes requested):
LGTM. Just left a request change to return list of Claims instead of list of pointers to Claims

internal/core/ports/claims_service.go
x1m3: I'm closing the comment as this change will affect a lot of places that should not be affected by this PR.

x1m3 (approved):
LGTM. Let's do the refactor proposed in a previous comment in other PR as it will add a lot of noise to this. Sorry

PR #596, opened 2024-01-12
feat: add qr code switcher to the issuer node

x1m3:
Ability to scan QR codes in modes: Link redirection, raw or base64 encoded.


ui/src/adapters/api/connections.ts
```diff
@@ -52,16 +57,18 @@ export async function getConnection({
 export async function getConnections({
   credentials,
   env,
-  params,
+  params: { max_results, page, query },
```
amonsosanz, line 60: We should apply also here what Elias mentioned in [this comment](https://github.com/0xPolygonID/issuer-node/pull/564#discussion_r1431197960).

ui/src/adapters/api/connections.ts
```diff
@@ -70,17 +77,22 @@ export async function getConnections({
...
     return buildSuccessResponse(
-      getListParser(connectionParser)
-        .transform(({ failed, successful }) => ({
-          failed,
-          successful: successful.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime()),
+      getResourceParser(connectionParser)
+        .transform(({ items: { failed, successful }, meta }) => ({
+          items: {
+            failed,
+            successful: successful.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime()),
```
amonsosanz, line 93: We should apply also here what Elias mentioned in [this comment](https://github.com/0xPolygonID/issuer-node/pull/564#discussion_r1431203737).

ui/src/components/connections/ConnectionsTable.tsx
```diff
@@ -14,6 +14,7 @@ import {
 import { useCallback, useEffect, useState } from "react";
 import { generatePath, useNavigate, useSearchParams } from "react-router-dom";
 
+import { positiveIntegerFromStringParser } from "../../adapters/parsers";
```
amonsosanz, line 17: We always use the absolute path on imports.

ui/src/components/credentials/CredentialIssuedQR.tsx
```diff
@@ -46,27 +48,30 @@ export function CredentialIssuedQR() {
...
+  if (!isAsyncTaskDataAvailable(issuedQRCodes)) {
     return <LoadingResult />;
   }
 
+  const [issuedQRCodeLink, issuedQRCodeRaw] = issuedQRCodes.data;
   return (
     <CredentialQR
-      qrCode={issuedQRCode.data.qrCodeLink}
-      schemaType={issuedQRCode.data.schemaType}
+      qrCodeBase64={`iden3comm://?i_m=${btoa(issuedQRCodeRaw.qrCode)}`}
```
amonsosanz, line 71: I'd remove this prop and move this logic to the component `CredentialQR`

ui/src/components/credentials/IssuanceMethodForm.tsx
```diff
@@ -52,11 +52,16 @@ export function IssuanceMethodForm({
 
   const fetchConnections = useCallback(
     async (signal: AbortSignal) => {
-      const response = await getConnections({ credentials: false, env, signal });
+      const response = await getConnections({
+        credentials: false,
+        env,
+        params: { max_results: 0 },
```
amonsosanz, line 58: Why this `max_results: 0`?

ui/src/adapters/api/connections.ts
```diff
@@ -4,10 +4,15 @@ import { z } from "zod";
 import { Response, buildErrorResponse, buildSuccessResponse } from "src/adapters";
```
amonsosanz, line 4: Can you please run `npm run lint:fix && npm run format` and commit the changes?

ui/src/adapters/api/connections.ts
x1m3: Done

ui/src/components/connections/ConnectionsTable.tsx
x1m3: Good catch. Changed

ui/src/components/credentials/CredentialIssuedQR.tsx
x1m3: I agree. Done,

ui/src/components/credentials/IssuanceMethodForm.tsx
x1m3: Removing it. Not needed. Probably I added it when debugging

x1m3: My intention was to download all connections from server to be shown in a  select. But, I think that this select is now an input text, so I think that this function has no sense right now. And can be a performance problem if the number of connections is too big.

Is this screen
imagen

amonsosanz: As [discussed in Slack](https://0xpolygon.slack.com/archives/C05AG4LR7ML/p1705679168364929?thread_ts=1705655683.487219&cid=C05AG4LR7ML) there's a bug in the API that returns `0` in `max_results`, making the response invalid, which means that this dropdown cannot be populated.

As agreed, we will set this page size (`max_results`) to `1000` and will sort the connection `DESC` so users will see the most recent `1000` connections listed, which we think covers most use cases.

ui/src/adapters/api/connections.ts
amonsosanz: mmm... I just realized we can indeed completely remove the `transform`. Would you please remove it also in `getCredentials`?

ui/src/components/credentials/IssuanceMethodForm.tsx
x1m3: Error fixed. Now server returns 1000 rows if parameter maxRows is not present.

ui/src/adapters/api/connections.ts
x1m3: sure. Removed!

amonsosanz: You removed the transform from `getLinks` not from `getCredentials`.

x1m3: Fixed. Transform restored in getLinks and removed from getCredentials

amonsosanz (approved):

PR #600, opened 2024-01-16
feat: adding Iden3commRevocationStatusV1

javip97:
16.1 -> Waiting for protocol implementation (mobile)


internal/config/credential_status.go
```diff
@@ -23,19 +25,24 @@ type CredentialStatus struct {
...
+	CredentialStatusType string `mapstructure:"CredentialStatusType" default:"Iden3commRevocationStatusV1"`
 }
 
 // DirectStatus is the type of direct status
 type DirectStatus struct {
 	URL string `mapstructure:"URL"`
 }
 
-// GetURL returns the URL of the direct status
+// GetURL returns the URL of the di	rect status
```
x1m3, line 36: Unwanted tab here

x1m3 (approved):
LGTM. Just an extra tab that we should remove.

internal/config/credential_status.go
javip97: Copilot :)

internal/core/services/revocation.go
```diff
@@ -106,6 +115,73 @@ func convertCredentialStatus(credStatus interface{}) (verifiable.CredentialStatu
 	return verifiable.CredentialStatus{}, errors.New("failed cast credential status to verifiable.CredentialStatus")
 }
 
+func getRevocationStatusFromAgent(ctx context.Context, from, to, endpoint string, nonce uint64) (*verifiable.RevocationStatus, error) {
```
vmidyllic, line 118: you should use this resolver
https://github.com/iden3/iden3comm/blob/main/resolvers/agent.go

as well as you can switch to Credential Registry concept here https://github.com/iden3/go-schema-processor/blob/adbd0a5aa31811583d88aa9f48ff00b4bb1efd9f/verifiable/credential_status.go#L33

internal/config/credential_status.go
```diff
@@ -36,6 +38,11 @@ func (r *DirectStatus) GetURL() string {
 	return strings.TrimSuffix(r.URL, "/")
 }
 
+// GetAgentURL returns the URL of the agent endpoint
+func (r *DirectStatus) GetAgentURL() string {
+	return fmt.Sprintf("%s/v1/agent", strings.TrimSuffix(r.URL, "/"))
```
vmidyllic, line 43: not sure if it is the right place

Direct status was about an endpoint, this is more like Iden3CommAgentStatus, but maybe it is ok.

internal/core/services/claims.go
```diff
@@ -597,6 +604,29 @@ func (c *claim) revoke(ctx context.Context, did *w3c.DID, nonce uint64, descript
...
 }
 
+func (c *claim) getRevocationStatus(ctx context.Context, basicMessage *ports.AgentRequest) (*domain.Agent, error) {
+	revData := &protocol.RevocationStatusRequestMessageBody{}
+	err := json.Unmarshal(basicMessage.Body, revData)
+	if err != nil {
+		return nil, fmt.Errorf("invalid revocation request body: %w", err)
+	}
+
+	var revStatus *verifiable.RevocationStatus
```
vmidyllic, line 614: you should reuse [this
](https://github.com/iden3/go-schema-processor/blob/adbd0a5aa31811583d88aa9f48ff00b4bb1efd9f/verifiable/credential_status.go#L33)

martinsaporiti: @vmidyllic I'm not sure if we reuse that. `ValidateCredentialStatus` function at the end calls the issuer node to get the credential rev status, and here we are resolving that. Maybe we can talk.

internal/config/credential_status.go
martinsaporiti: I've changed this. It was wrong here

internal/core/services/revocation.go
martinsaporiti: this was deleted because It was not used

internal/config/credential_status.go
```diff
@@ -6,6 +6,7 @@ import (
 
 const (
 	sparseMerkleTreeProof                 = "SparseMerkleTreeProof"
+	iden3commRevocationStatusV1           = "Iden3commRevocationStatusV1.0"
 	iden3ReverseSparseMerkleTreeProof     = "Iden3ReverseSparseMerkleTreeProof"
 	iden3OnchainSparseMerkleTreeProof2023 = "Iden3OnchainSparseMerkleTreeProof2023"
```
vmidyllic, line 11: you can reuse these constants from libraries

pkg/credentials/revocation_status/iden3_revocation_status_v1.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/iden3/go-schema-processor/v2/verifiable"
+
+	"github.com/polygonid/sh-id-platform/internal/config"
+)
+
+type iden3CommRevocationStatusV1Resolver struct{}
+
+func (r *iden3CommRevocationStatusV1Resolver) resolve(credentialStatusSettings config.CredentialStatus, issuerDID w3c.DID, nonce uint64, issuerState string) *verifiable.CredentialStatus {
+	return &verifiable.CredentialStatus{
+		ID:              buildDirectRevocationURL(credentialStatusSettings.Iden3CommAgentStatus.GetURL()),
```
vmidyllic, line 14: agent url must be enough, why do you need to build direct url here?

pkg/credentials/revocation_status/sparse_merkletree_proof_resolver.go
```diff
@@ -9,9 +9,9 @@ import (
 
 type sparseMerkleTreeProofResolver struct{}
 
-func (r *sparseMerkleTreeProofResolver) resolve(credentialStatusSettings config.CredentialStatus, issuerDID w3c.DID, nonce uint64, issuerState string) *verifiable.CredentialStatus {
+func (r *sparseMerkleTreeProofResolver) resolve(credentialStatusSettings config.CredentialStatus, _ w3c.DID, nonce uint64, _ string) *verifiable.CredentialStatus {
 	return &verifiable.CredentialStatus{
-		ID:              buildDirectRevocationURL(credentialStatusSettings.DirectStatus.GetURL(), issuerDID.String(), nonce, credentialStatusSettings.SingleIssuer),
+		ID:              buildDirectRevocationURL(credentialStatusSettings.Iden3CommAgentStatus.GetURL()),
```
vmidyllic, line 14: smt status  (direct) is a separate, if you don't need it on the issuer node, maybe you don't need this file? but such fix looks strange.

pkg/credentials/revocation_status/utils.go
```diff
@@ -15,12 +14,8 @@ func buildRHSRevocationURL(host, issuerState string) string {
 	return fmt.Sprintf("%s/node?state=%s", host, issuerState)
 }
 
-func buildDirectRevocationURL(host, issuerDID string, nonce uint64, singleIssuer bool) string {
-	if singleIssuer {
-		return fmt.Sprintf("%s/v1/credentials/revocation/status/%d",
-			host, nonce)
-	}
-	return fmt.Sprintf("%s/v1/%s/claims/revocation/status/%d", host, url.QueryEscape(issuerDID), nonce)
+func buildDirectRevocationURL(host string) string {
```
vmidyllic, line 17: I would rename buildDirectRevocationURL to build agent url, not confuse developers in other places.

pkg/protocol/packagemanager.go
```diff
@@ -1,68 +1,35 @@
...
-
-	zkpPackerV2 := packers.NewZKPPacker(
-		provers,
-		verifications,
-	)
-
-	// TODO: Why jwsPacker is not defined here?
+	verifications[jwz.AuthV2Groth16Alg] = packers.NewVerificationParams(authV2Set.VerificationKey, stateVerificationHandler(stateContract))
 
+	zkpPackerV2 := packers.NewZKPPacker(nil, verifications)
```
vmidyllic, line 27: are you sure you don't need prover here?

pkg/credentials/revocation_status/sparse_merkletree_proof_resolver.go
martinsaporiti: that's correct, We don't need the file.

pkg/credentials/revocation_status/utils.go
martinsaporiti: I will remove the function because it is no longer needed

pkg/protocol/packagemanager.go
martinsaporiti: I've found the prover configured here is not being used

PR #604, opened 2024-01-18
feat: UI can create credentials with refresh service field

x1m3:
Changes in Create/Show Credential interface:
- There is a new field refresh service to be included in the credential. This field needs the credential to have an expiration date.
- Credential expiration time becomes a time. Not only a date. This is useful for credentials with short live time.


ui/src/adapters/parsers/forms.ts
```diff
@@ -126,6 +129,18 @@ const issueCredentialFormDataParser = getStrictParser<IssueCredentialFormData>()
...
       .min(1, "At least one proof type is required"),
+    refreshService: z.union([
+      z.object({
+        enabled: z.literal(false),
+        url: z.string(),
+      }),
+      z.object({
+        enabled: z.literal(true),
+        url: z.string().url({
+          message: `Refresh service URL must be a valid URL.`,
```
amonsosanz, line 140: I'd suggest using `URL_FIELD_ERROR_MESSAGE` here.

ui/src/adapters/parsers/forms.ts
```diff
@@ -126,6 +129,18 @@ const issueCredentialFormDataParser = getStrictParser<IssueCredentialFormData>()
     proofTypes: z
       .array(z.union([z.literal("MTP"), z.literal("SIG")]))
       .min(1, "At least one proof type is required"),
+    refreshService: z.union([
+      z.object({
+        enabled: z.literal(false),
+        url: z.string(),
```
amonsosanz, line 135: Why do we need `url` as string when `enabled` is false?

ui/src/adapters/parsers/forms.ts
```diff
@@ -145,11 +160,21 @@ export const credentialFormParser = getStrictParser<
...
+        issueCredential;
       const { type } = issuanceMethod;
 
+      if (refreshService.enabled && credentialExpiration === undefined) {
+        context.addIssue({
+          code: z.ZodIssueCode.custom,
+          fatal: true,
+          message: `Credential expiration must set when refresh service is enabled.`,
+        });
+      }
```
amonsosanz, line 173: The fact that you need to add an error in the transformation phase just by checking the values (not transforming them in a way that can fail) is an indicator that the parser is incorrect.

`transform` is a map function that is called only when the parsing succeeds and when the parsing succeeds no further checking on the values should be required.

ui/src/adapters/parsers/forms.ts
```diff
@@ -319,13 +345,17 @@ export function serializeCredentialLinkIssuance({
   if (serializedSchemaForm.success) {
     return {
       data: {
-        credentialExpiration: credentialExpiration
-          ? serializeDate(credentialExpiration, "date")
-          : null,
+        credentialExpiration: credentialExpiration ? credentialExpiration.toISOString() : null,
```
amonsosanz, line 348: We should keep using `serializeDate` and just pass the right output format, in this case, `"date-time"`.

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -216,6 +224,11 @@ export function CredentialDetails() {
                       text={expiresAt ? formatDate(expiresAt) : "-"}
                     />
 
+                    <Detail
+                      label="Refresh Service"
+                      text={refreshService ? " " + refreshService.id : "-"}
```
amonsosanz, line 229: Instead of doing this, we should adapt the `Detail` component to cover this new requirement.

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -41,6 +41,7 @@ const defaultCredentialFormInput: CredentialFormInput = {
   },
   issueCredential: {
     proofTypes: ["SIG"],
+    refreshService: { enabled: false, url: "" },
```
amonsosanz, line 44: `url` should not be required when enabled is false.

ui/src/domain/credential.ts
```diff
@@ -2,13 +2,19 @@ export type CredentialsTabIDs = "issued" | "links";
 
 export type ProofType = "MTP" | "SIG";
 
+export type RefreshService = {
+  id: string;
+  type: string;
```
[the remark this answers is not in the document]
x1m3: By now, "Iden3RefreshService2023" is the only refresh service version availiable

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -437,9 +441,35 @@ export function IssueCredentialForm({
...
+                          <Form.Item
+                            name={["refreshService", "enabled"]}
+                            noStyle
+                            valuePropName="checked"
+                          >
+                            <Checkbox
+                              checked={refreshServiceChecked}
+                              onChange={() => {
+                                setRefreshServiceChecked(!refreshServiceChecked);
+                              }}
```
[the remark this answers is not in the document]
x1m3: The proposed sollution is not working for me. It seems that the state is not properly managed. Tested in firefox. Need to work on this.

ui/src/adapters/api/credentials.ts
```diff
@@ -130,6 +131,10 @@ export type CreateCredential = {
   credentialSubject: Json;
   expiration: string | null;
   mtProof: boolean;
+  refreshService: {
+    id: string;
+    type: "Iden3RefreshService2023";
+  } | null;
```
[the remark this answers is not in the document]
x1m3: Done. I didn't wanted to couple the api with the domain, but as I see it is common rule and not a risk, I accept the proposal.

ui/src/adapters/parsers/forms.ts
x1m3: Because it needs to adhere to the IssueCredentialFormData type that is how the form is defined.

ui/src/adapters/parsers/forms.ts
x1m3: Agree. Validation moved to form level.

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -437,9 +441,35 @@ export function IssueCredentialForm({
...
+                            >
+                              Enable
+                            </Checkbox>
+                          </Form.Item>
+                          <Form.Item name={["refreshService", "url"]} noStyle>
```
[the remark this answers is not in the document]
x1m3: Better solution. Thanks. Fixed.

ui/src/adapters/parsers/forms.ts
x1m3: Yes. Change reverted. I also change some other places where to.ISOString() was being used.

api_ui/api.yaml
```diff
@@ -1518,6 +1518,7 @@ components:
         # refresh service
     RefreshService:
       type: object
+      x-omitempty: false
```
mabaro3009, line 1521: test @x1m3

ui/src/components/credentials/CredentialDetails.tsx
x1m3: Fixed. I've updated the <Detail> component adding an href parameter. If this parameter is present, the <Detail> will convert that text into a link like <a href="{href}">{text}</a>

ui/src/components/credentials/IssueCredential.tsx
x1m3: Please, see previous comment https://github.com/0xPolygonID/issuer-node/pull/604#discussion_r1464608691 I do not know how to make url nullable. But I don't see a risk in having a default value of empty string.

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -437,9 +446,59 @@ export function IssueCredentialForm({
...
+                          </Form.Item>
+                        </Space>
+                      </Form.Item>
+                      <Form.Item
+                        label="Credential expiration date"
+                        name="credentialExpiration"
+                        rules={[
+                          {
+                            message:
+                              "Credential expiration must set when refresh service is enabled.",
```
amonsosanz, line 490: `Credential expiration must be set when the refresh service is enabled`

ui/src/components/credentials/IssueCredentialForm.tsx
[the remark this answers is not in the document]
amonsosanz: You're right. I missed this warning:

image

The form is not yet mounted when we try to access its state. Let's leave the `refreshServiceChecked` state as you have it.

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -216,6 +228,11 @@ export function CredentialDetails() {
                       text={expiresAt ? formatDate(expiresAt) : "-"}
                     />
 
+                    <Detail
+                      label="Refresh Service"
+                      text={refreshService ? refreshService.id : "-"}
```
amonsosanz, line 233: Aren't we missing the `href` here?

I also think you're missing it in `ImportSchemaPreview.tsx`, `ImportSchemaPreview.tsx` and `SchemaDetails.tsx`.

ui/src/adapters/parsers/forms.ts
x1m3: Good suggestion.  Changed

amonsosanz (approved):

PR #614, opened 2024-01-30
feat: Implement remote sorting for Credentials and Connections

amonsosanz:
This PR implements remote sorting for the tables Credentials and Connections.

### DIFF
--- a/ui/src/adapters/api/index.ts
+++ b/ui/src/adapters/api/index.ts
@@ -3,6 +3,18 @@ import { z } from "zod";
 import { getStrictParser } from "src/adapters/parsers";
 import { Env } from "src/domain";
 
+export type Sorter = { field: string; order?: "ascend" | "descend" };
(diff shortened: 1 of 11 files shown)


ui/src/adapters/api/index.ts
```diff
@@ -3,6 +3,21 @@ import { z } from "zod";
 import { getStrictParser } from "src/adapters/parsers";
 import { Env } from "src/domain";
 
+export type Sorter = { field: string; order: "ascend" | "descend" };
```
elias-garcia, line 6: "ascend" and "descend" looks weird as they are verbs. Order is "ascendant" or "descendant".

amonsosanz: [It's what it's...](https://ant.design/components/table#api)
image

elias-garcia (approved):

PR #624, opened 2024-02-28
fix: Pagination in connections 

x1m3:
Pagination in connections was failing due to a some incompatibilities with tables joins and offset limit implementation.


martinsaporiti (approved):
LGMT

javip97 (approved):

PR #642, opened 2024-04-01
add separate error variable for setLink()

yushihang:
If only one error is used, when the first value of "err" is not nil and second value of "err" is nil, there may be a situation that "the function execution should marked as failed, but it returns nil"


x1m3:
Thanks for your contribution @yushihang . I'm going to change the destination branch to develop. 

Your suggestions seems very good to me

OBrezhniev (approved):

x1m3 (approved):

PR #644, opened 2024-04-05
fix: update identity state for mtp credentials

martinsaporiti:


internal/core/services/identity.go
```diff
@@ -310,6 +310,7 @@ func (i *identity) GetKeyIDFromAuthClaim(ctx context.Context, authClaim *domain.
 }
 
 func (i *identity) UpdateState(ctx context.Context, did w3c.DID) (*domain.IdentityState, error) {
+	const authBJJCredential = "https://schema.iden3.io/core/jsonld/auth.jsonld#AuthBJJCredential"
```
x1m3, line 313: I think that we could use domain.AuthBJJCredentialTypeID constant

x1m3 (commented):
LGTM. Just left a comment regarding a constant that probably we could reuse

x1m3 (approved):

javip97 (approved):
LGTM

x1m3 (approved):
LGTM

PR #645, opened 2024-04-05
feat: adding amoy support

javip97:
Adding Amoy support to the Issuer Node.

It sets Amoy as the default testing network due to the deprecation of Mumbai.

More info about Amoy: https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos


martinsaporiti (approved):
LGTM

PR #646, opened 2024-04-08
fix: publish state and improve performance when processing takes place

martinsaporiti:
## Notes
* Fix and improve the publish state process.

### DIFF
--- a/internal/core/services/identity.go
+++ b/internal/core/services/identity.go
@@ -56,7 +56,7 @@ var (
 	// ErrAssigningMTPProof - represents an error in the identity metadata
 	ErrAssigningMTPProof = errors.New("error assigning the MTP Proof from Auth Claim. If this identity has keyType=ETH you must to publish the state first")
 	// ErrNoClaimsFoundToProcess - means that there are no claims to process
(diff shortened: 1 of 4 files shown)


internal/core/services/identity.go
```diff
@@ -403,17 +396,28 @@ func (i *identity) UpdateState(ctx context.Context, did w3c.DID) (*domain.Identi
...
+
+	if claimsAddedToTree {
+		for i := range lc {
+			err = iTrees.AddClaim(ctx, &lc[i])
+			if err != nil {
+				log.Error(ctx, "adding claim to tree", "err", err)
+				return false, err
+			}
+		}
+	}
```
x1m3, line 419: Not sure if all this could be simplified with just a 
```
for i := range lc {
	err = iTrees.AddClaim(ctx, &lc[i])
	if err != nil {
		log.Error(ctx, "adding claim to tree", "err", err)
		return false, err
	}
}
return len(lc)>0
```

x1m3 (approved):
LGTM

javip97 (approved):

x1m3 (approved):

PR #647, opened 2024-04-09
feat: fuzzy query all the content in the value of credentialSubject

jduan1993:
- Used to fuzzy query all the content in the value of credentialSubject, case insensitive.
- It can be applied to scenarios where did owner needs to be quickly located. The owner information is contained in the field of credentialSubject, and the fuzzy query can be used to locate the credential to realize the corresponding operation, such as revocation.
- Example: For example, if the account of a natural person needs to be restricted or closed, we can use this feature.

### DIFF
--- a/internal/api_ui/api.gen.go
+++ b/internal/api_ui/api.gen.go
@@ -137,15 +137,15 @@ type CreateCredentialRequest struct {
 
 // CreateLinkRequest defines model for CreateLinkRequest.
 type CreateLinkRequest struct {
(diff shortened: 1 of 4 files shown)


x1m3:
Please @jduan1993 . Can you change the base destination to develop brach?  Our usual flow is adding new functionalities into develop branch and every one or two months merging develop to main and creating a new release. 

This is explained in the CONTRIBUTING document https://github.com/0xPolygonID/issuer-node/blob/develop/CONTRIBUTING.md  (unfortunately this doc is still on develop branch because it was part of current release)

PR #656, opened 2024-04-16
Add config gasless network

daveroga:
When integrating Issuer Node with Polygon CDK or other gasless networks there are some issues in sending transactions to the RPC node regarding the estimation of the gas.

Add boolean parameter ISSUER_ETHEREUM_GASLESS that will take into account to calculate or not the estimation of the gas and the parameters needed for sending the transaction properly to the RPC node.


pkg/blockchain/eth/client.go
```diff
@@ -357,63 +365,72 @@ func (c *Client) CreateRawTx(ctx context.Context, txParams TransactionParams) (*
...
+			txParams.BaseFee = baseFee
+		}
+
+		if txParams.GasTips == nil {
+			_ctx3, cancel3 := context.WithTimeout(ctx, c.Config.RPCResponseTimeout)
+			defer cancel3()
+			gasTip, err := c.client.SuggestGasTipCap(_ctx3)
+			// since hardhad doesn't support 'eth_maxPriorityFeePerGas' rpc call.
+			// we should hardcode 0 as a mainer tips. More information: https://github.com/NomicFoundation/hardhat/issues/1664#issuecomment-1149006010
+			if err != nil && strings.Contains(err.Error(), "eth_maxPriorityFeePerGas not found") {
```
martinsaporiti, line 420: Maybe we can use something like this:
    
```
if errors.Is(err, erthMexPriorityFeePerGasErr) {
        ...
    }
```
and change the error returned by the client

daveroga: I didn't touch this piece of code in this PR. It comes from previous code.

martinsaporiti (approved):
LGTM

PR #671, opened 2024-06-13
fix: invisible characters

roderik:
There are some invisible characters in this value, which causes the url to have `%E2%80%8D` after amoy


x1m3 (approved):

x1m3:
Thanks for your PR @roderik

PR #673, opened 2024-06-14
chore: add ui insecure option

martinsaporiti:
* Add the option to make issuer node UI insecure (credentials are not needed)


x1m3 (approved):
LGTM

PR #674, opened 2024-06-17
fix: verify iden3comm message media type

ilya-korotya:
… media type


internal/core/services/claims.go
```diff
@@ -365,7 +368,13 @@ func (c *claim) GetCredentialQrCode(ctx context.Context, issID *w3c.DID, id uuid
 	}, nil
 }
 
-func (c *claim) Agent(ctx context.Context, req *ports.AgentRequest) (*domain.Agent, error) {
+func (c *claim) Agent(ctx context.Context, req *ports.AgentRequest, mediatype iden3comm.MediaType) (*domain.Agent, error) {
+	if !c.mediatypeManager.AllowMediaType(req.Type, mediatype) {
+		err := fmt.Errorf("unsupported media type '%s' for message type '%s'", req.Typ, req.Type)
+		log.Error(ctx, "err", err)
```
x1m3, line 374:
```suggestion
		log.Error(ctx, "agent: unsupported media type", err", err)
```

internal/core/services/identity.go
```diff
@@ -735,30 +737,44 @@ func (i *identity) createIdentity(ctx context.Context, tx db.Querier, hostURL st
...
-						&merkletree.HashZero,
-						&merkletree.HashZero,
+		if rhsMode == reverse_hash.RHSModeOnChain && !i.autoPublishingToOnChainRHS {
+			log.Info(ctx, `
+			state info (hashes):
+				root: '%s',
+				ClaimsTreeRoot: '%s',
+				RevocationsTreeRoot: '%s',
+				RootOfRootsTreeRoot: '%s',
+		`, identity.State.TreeState().State, claimsTree.Root(), &merkletree.HashZero, &merkletree.HashZero)
```
x1m3, line 760: I guess that this log is not working. Probably it should be a log.Debug.  And the signature for issuer node log is 
log(ctx, "message", "key1", v1, "key2", v2,....) And usually logs are written in a single line because they are not relevant for the reader.

```suggestion
			log.Info(ctx, "state info (hashes):","root",  '%s', identity.State.TreeState().State, "ClaimsTreeRoot", claimsTree.Root()"... etc..
```

internal/core/services/mediatype_manager.go
```diff
@@ -0,0 +1,48 @@
...
+		strictMode: strictMode,
+		allowList:  allowList,
+	}
+}
+
+// AllowMediaType check if the protocol message supports the mediaType type
+func (m *MediaTypeManager) AllowMediaType(
+	protoclMessage iden3comm.ProtocolMessage,
+	mediaType iden3comm.MediaType,
+) bool {
```
x1m3, line 37: Usually,all funcion signatures in this project are in one line, except some log exceptions

x1m3 (commented):
LGTM. I just left some small style suggestions.  Please, find a good name for the PR so we will have a nice description in changelog ;-)

x1m3 (approved):
LGTM
We should document this new value

PR #677, opened 2024-06-21
PID-2239: add privado-chain

volodymyr-basiuk:
- updated go-libs with privado-main and privado-test network provided by default

NOTE: please remove `ISSUER_CUSTOM_DID_METHODS` from `.env-issuer` while deploying this changes


pkg/reverse_hash/factory.go
```diff
@@ -85,7 +85,7 @@ func (f *Factory) initOnChainRHSCli(ctx context.Context, kmsKey *kms.KeyID) (pro
 	if err != nil {
 		return nil, err
 	}
-	cli, err := proofEth.NewReverseHashCli(f.onChainTreeStoreContract, f.ethClient.GetEthereumClient(), txOpts, f.responseTimeout)
+	cli, err := proofEth.NewReverseHashCli(f.ethClient.GetEthereumClient(), f.onChainTreeStoreContract, txOpts.From, txOpts.Signer)
```
vmidyllic, line 88: @volodymyr-basiuk is it an only thing to change to support new version of onchain rhs cli?
@martinsaporiti please double check that onchain rhs is working on staging / test

vmidyllic (approved):

pkg/reverse_hash/factory.go
volodymyr-basiuk: I have checked with `ISSUER_CREDENTIAL_STATUS_RHS_MODE=OnChain`, it works

x1m3 (approved):
LGTM

PR #678, opened 2024-06-21
chore: add custom did methods

martinsaporiti:
* Add support for custom did methods in Helm chart
* improve service account definition


x1m3 (approved):
LGTM

PR #681, opened 2024-06-25
chore: add proofs to create claim endpoint

martinsaporiti:
## Notes
* Add proofs field to the create claim body endpoint in core API.
* The values allowed are:  BJJSignature2021┃Iden3SparseMerkleTreeProof

With this feature the user can specify the type of proof for the claim

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -151,6 +151,9 @@ func (s *Server) CreateIdentity(ctx context.Context, request CreateIdentityReque
 
 // CreateClaim is claim creation controller
 func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObject) (CreateClaimResponseObject, error) {
+	const BJJSignatureProof2021 = "BJJSignatureProof2021"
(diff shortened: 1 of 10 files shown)


internal/api/server.go
```diff
@@ -151,6 +151,9 @@ func (s *Server) CreateIdentity(ctx context.Context, request CreateIdentityReque
 
 // CreateClaim is claim creation controller
 func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObject) (CreateClaimResponseObject, error) {
+	const BJJSignatureProof2021 = "BJJSignatureProof2021"
```
OBrezhniev, line 154: Can you use constants from here: https://github.com/iden3/go-schema-processor/blob/684a4b617a0eaf8452d4f3a65a4a6704ad53cbdf/verifiable/constants.go#L40

OBrezhniev: Or the ones you defined in internal/api/api.gen.go

OBrezhniev (approved):

PR #686, opened 2024-07-05
Pid 2289 support jws packer

martinsaporiti:


vmidyllic:
do we really need to support signer with bjj on the issuer node?

martinsaporiti:
> do we really need to support signer with bjj on the issuer node?

good catch, thanks

vmidyllic (approved):

cmd/platform/main.go
```diff
@@ -154,7 +154,7 @@ func main() {
 	}
 
 	publisher := gateways.NewPublisher(storage, identityService, claimsService, mtService, keyStore, transactionService, proofService, publisherGateway, networkResolver, ps)
-	packageManager, err := protocol.InitPackageManager(ctx, networkResolver.GetSupportedContracts(), cfg.Circuit.Path)
+	packageManager, err := protocol.InitPackageManager(ctx, networkResolver.GetSupportedContracts(), cfg.Circuit.Path, auth.UniversalDIDResolver)
```
vmidyllic, line 157: it is better to provide an ability to configure and pass it both to auth verifier and here, so production will not rely on dev instance of third party project. (as fallback it is ok)

vmidyllic (changes requested):

x1m3 (approved):
LGTM

cmd/platform/main.go
martinsaporiti: solved

vmidyllic (approved):

PR #687, opened 2024-07-10
Pid 2179 add support aws kms for key management

martinsaporiti:


cmd/kms_priv_key_importer/readme.md
```diff
@@ -0,0 +1,113 @@
...
+```logs
+2024/07/10 10:38:14 INFO alias created: alias:=alias/pbkey
+2024/07/10 10:38:14 INFO key created keyId=157a8b2a-e5e9-4414-b9c5-301ce828f6c5
+```
+
+then you can import your private key using the following command:
+
+```shell
+$ chmod +x kms_priv_key_importer
+$ ./kms_priv_key_importer <privateETHKey> <key-id> <aws-profile>
```
ilya-korotya, line 69: If I correctly understand the flow. It should be `aws_kms_material_key_imporer.sh` not the golang bin file.

cmd/kms_priv_key_importer/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+if you get `Key material successfully imported!!!` message, then your private key was successfully imported to AWS KMS.
+
+### Running Importer with Docker (AWS)
+In the root project folder run:
+
+```shell
+docker build --build-arg ISSUER_KMS_ETH_PLUGIN_AWS_ACCESS_KEY=XXXX \
+  --build-arg ISSUER_KMS_ETH_PLUGIN_AWS_SECRET_KEY=YYYY \
+  --build-arg ISSUER_KMS_ETH_PLUGIN_AWS_REGION=eu-west-1 -t privadoid-kms-importer -f Dockerfile-kms-importer .
```
ilya-korotya, line 86: All other paths to files in this readme.md link to `.` (current directory). But `Dockerfile-kms-importer` exists on the root directory. Change to `../Dockerfile-kms-importer`

cmd/kms_priv_key_importer/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+```shell
+2024/07/10 15:27:54 INFO alias created: alias:=alias/pbkey
+2024/07/10 15:27:54 INFO key created keyId=9bb5b78b-c288-44a7-b1d4-0543e0a6
+```
+
+then import the material key
+
+```shell
+sh ./aws_kms_material_key_importer.sh ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 9bb5b78b-c288-44a7-b1d4-0543e0a6 privadoid
```
ilya-korotya, line 111: `ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80` is a real private key with some balance. Please creat an empty one.

internal/providers/vault.go
```diff
@@ -71,40 +73,56 @@ func VaultClient(ctx context.Context, cfg Config) (*vault.Client, error) {
...
 
 // newVaultClientWithToken checks vault configuration and creates new vault client
-func newVaultClientWithToken(address, token string) (*vault.Client, error) {
-	if address == "" {
+func newVaultClientWithToken(cfg Config) (*vault.Client, error) {
+	if cfg.Address == "" {
 		return nil, errors.New("vault address is not specified")
 	}
-	if token == "" {
+	if cfg.Address == "" {
```
ilya-korotya, line 80: Should be `cfg.Token`. Because we check Address field on line 77

.env-issuer.sample
```diff
@@ -12,22 +13,47 @@ ISSUER_LOG_LEVEL=-4
...
-ISSUER_KEY_STORE_ADDRESS=http://vault:8200
-ISSUER_KEY_STORE_PLUGIN_IDEN3_MOUNT_PATH=iden3
-# Could be either localstorage or vault
-ISSUER_KMS_PLUGIN=localstorage
+
+# --------------------------------------------------------------------------------
+# KMS configuration
+# --------------------------------------------------------------------------------
+# Could be either [localstorage | vault] (BJJ) and [localstorage | vault | aws ] (ETH)
+ISSUER_KMS_BJJ_PLUGIN=vault
```
ilya-korotya, line 21: In my opinion, the name `ISSUER_KMS_BJJ_PROVIDER` is clearer. A plugin is something that is embedded in existing code (like custom plugins for vault) otherwise, the provider is a separate service (vault, was kms) that provides some service.  I think a provider is a better variant.

ilya-korotya (changes requested):
1. I noticed that in some places we are using `cors.Handler(cors.Options{AllowedOrigins: []string{"*"}})` it is very bad. We should pass all possible originals from config.
2. If I understood the idea correctly. If I want to migrate a key to a vault or local storage, I should use a Makefile, but if I want to migrate a key to AWS KMS, I should run `aws_kms_material_key_importer.sh`. Why can't we have a generic Makefile command to move the key anywhere?

cmd/kms_priv_key_importer/readme.md
martinsaporiti: the readme specifies "In the root project folder run:"

martinsaporiti:
> 1. I noticed that in some places we are using `cors.Handler(cors.Options{AllowedOrigins: []string{"*"}})` it is very bad. We should pass all possible originals from config.
> 2. If I understood the idea correctly. If I want to migrate a key to a vault or local storage, I should use a Makefile, but if I want to migrate a key to AWS KMS, I should run `aws_kms_material_key_importer.sh`. Why can't we have a generic Makefile command to move the key anywhere?



> 1. I noticed that in some places we are using `cors.Handler(cors.Options{AllowedOrigins: []string{"*"}})` it is very bad. We should pass all possible originals from config.
> 2. If I understood the idea correctly. If I want to migrate a key to a vault or local storage, I should use a Makefile, but if I want to migrate a key to AWS KMS, I should run `aws_kms_material_key_importer.sh`. Why can't we have a generic Makefile command to move the key anywhere?

Regarding 1, that was corrected in another PR.
Regarding 2, first of all it is not a migration, it is an import process. This import process requires different steps depending on where you want to import the eth private key. To be able to do it in a single makefile command, you would have to make an additional sh file that includes the possible options, and then it becomes more difficult. to detect if something fails, where the problem is. I don't see much benefit either from having a make command to run a sh file.

ilya-korotya:
I'm okay with point 1. 
Point 2: I think your idea with two different points to import keys is OK. But in my opinion better to have one golang binary that can import keys to any place. One common binary is easier to support than a bash script. The Golang bin on docker is OS agnostic. For bash script different Linux dist. can have different ports of open SSL and etc.

martinsaporiti:
> I'm okay with point 1. Point 2: I think your idea with two different points to import keys is OK. But in my opinion better to have one golang binary that can import keys to any place. One common binary is easier to support than a bash script. The Golang bin on docker is OS agnostic. For bash script different Linux dist. can have different ports of open SSL and etc.

Doing with Golang the same we are doing with the script is not trivial and after a lot of time trying to do it I think it's not possible. But if you want to try, you are welcome. I  tried different approaches without success.

Makefile
```diff
@@ -159,16 +162,46 @@ add-private-key:
...
+## AWS: make private_key=XXX aws_access_key=YYY aws_secret_key=ZZZ aws_region=your-region import-private-key-to-kms
+## localstorage and vault: make private_key=XXX import-private-key-to-kms
+.PHONY: import-private-key-to-kms
+import-private-key-to-kms:
+ifeq ($(ISSUER_KMS_ETH_PROVIDER), aws)
+	@echo "AWS"
+	docker build --build-arg ISSUER_KMS_ETH_PROVIDER_AWS_ACCESS_KEY=$(aws_access_key) \
+    		  --build-arg ISSUER_KMS_ETH_PROVIDER_AWS_SECRET_KEY=$(aws_secret_key) \
+    		  --build-arg ISSUER_KMS_ETH_PROVIDER_AWS_REGION=$(aws_region) -t privadoid-kms-importer -f ./Dockerfile-kms-importer .
+	$(eval result = $(shell docker run -it -v ./.env-issuer:/.env-issuer -v $(ISSUER_KMS_PROVIDER_LOCAL_STORAGE_FILE_PATH)/kms_localstorage_keys.json:/localstoragekeys/kms_localstorage_keys.json \
```
ilya-korotya, line 175: Why do we need to pass the path `/localstoragekeys/kms_localstorage_keys.json` to localstorage_key if we import a key to AWS?

ilya-korotya (approved):

PR #688, opened 2024-07-12
chore: add id parameter when claim is created

martinsaporiti:


x1m3 (approved):
I'm not a fan of injecting external ids to a system, but I guess the DB unique index constraint will protect us

PR #691, opened 2024-07-17
chore: Organize code in shorter files.

x1m3:


internal/api/credentials.go
```diff
@@ -0,0 +1,522 @@
...
+// CreateClaim is claim creation controller
+// deprecated - Use CreateCredential instead
+func (s *Server) CreateClaim(ctx context.Context, request CreateClaimRequestObject) (CreateClaimResponseObject, error) {
+	did, err := w3c.ParseDID(request.Identifier)
+	if err != nil {
+		return CreateClaim400JSONResponse{N400JSONResponse{Message: err.Error()}}, nil
+	}
+	var expiration *time.Time
+	if request.Body.Expiration != nil {
+		expiration = common.ToPointer(time.Unix(*request.Body.Expiration, 0))
```
ilya-korotya, line 55: I'm not sure, but maybe we should use UTC? `common.ToPointer(time.Unix(*request.Body.Expiration, 0).UTC())`
There is an example with UTC: https://pkg.go.dev/time#Unix

internal/api/identity.go
```diff
@@ -0,0 +1,217 @@
...
+
+		if errors.Is(err, kms.ErrPermissionDenied) {
+			var message string
+			if s.cfg.VaultUserPassAuthEnabled {
+				message = "Issuer Node cannot connect with Vault. Please check the value of ISSUER_VAULT_USERPASS_AUTH_PASSWORD variable."
+			} else {
+				message = `Issuer Node cannot connect with Vault. Please check the value of ISSUER_KEY_STORE_TOKEN variable.`
+			}
+
+			log.Info(ctx, message+". More information in this link: https://devs.polygonid.com/docs/issuer/vault-auth")
```
ilya-korotya, line 63: Link to PrivadoID?

internal/api/identity.go
```diff
@@ -0,0 +1,217 @@
...
+
+	if keyType != "BJJ" && keyType != "ETH" {
+		return CreateIdentity400JSONResponse{
+			N400JSONResponse{
+				Message: "Type must be BJJ or ETH",
+			},
+		}, nil
+	}
+
+	rhsSettings, err := s.networkResolver.GetRhsSettingsForBlockchainAndNetwork(ctx, blockchain, network)
```
ilya-korotya, line 34: I little bit confused with this part. Need some time to understand it.

x1m3:
Well. This PR is only moving code from one file to another. I assume that, in the scope of this PR, code works.

internal/api/identity.go
ilya-korotya: I was confused about the naming. The method `GetRhsSettingsForBlockchainAndNetwork` should be renamed to `GetRevocationSettingsForBlockchainAndNetwork` because this method also returns information for other revocation statuses, not just the RHS revocation status.

Secondly, I think I missed this earlier, but we should change this (not in this PR). Why do we define the revocation status not for the issuer but for the chain? On the old issuer server, we had different logic to manage different revocation statuses. On the old issuer server, it was possible to create an issuer and specify the type of revocation for the issuer:
POST /identitye/create
```json
{
  blockchain:
  method:
  network:
  type:
authBJJCredentialStatus: <where we can set one value from the list: Iden3ReverseSparseMerkleTreeProof|SparseMerkleTreeProof|...|etc"
}
```

ilya-korotya (approved):

internal/api/credentials.go
x1m3: I don't think so. Using UTC just change the time representation to be in UTC zone, but time is the same. The time.Unix() works in UTC

internal/api/identity.go
x1m3: Good point. Let's not forget about this and find a time to discuss it. @martinsaporiti FYI.

PR #693, opened 2024-07-21
feat: Add Mode All and support credential statuses on API level

volodymyr-basiuk:
- Added a new `Mode` option - `All` (publish to both OnChain and OffChain RHS).
- For the `Create Identity` endpoint, added an optional `authBJJCredentialStatus` parameter (if omitted, it creates an identity with Iden3commRevocationStatusV1.0).
- For the `Create Credential` and `CreateLinkQrCode` endpoints, added a `credentialStatusType` parameter (if omitted, it uses `Iden3commRevocationStatusV1.0` status type).


internal/api/identity.go
```diff
@@ -31,17 +33,28 @@ func (s *Server) CreateIdentity(ctx context.Context, request CreateIdentityReque
...
+	} else {
+		authBJJCredentialStatus = verifiable.Iden3commRevocationStatusV1
+	}
+
 	rhsSettings, err := s.networkResolver.GetRhsSettingsForBlockchainAndNetwork(ctx, blockchain, network)
 	if err != nil {
 		return CreateIdentity400JSONResponse{N400JSONResponse{Message: fmt.Sprintf("error getting reverse hash service settings: %s", err.Error())}}, nil
 	}
 
+	if !s.networkResolver.IsCredentialStatusTypeSupported(rhsSettings, authBJJCredentialStatus) {
```
x1m3, line 48: log please.

internal/api/identity.go
```diff
@@ -31,17 +33,28 @@ func (s *Server) CreateIdentity(ctx context.Context, request CreateIdentityReque
 		}, nil
 	}
 
+	var authBJJCredentialStatus verifiable.CredentialStatusType
+	if authBJJCredentialStatusString != nil && *authBJJCredentialStatusString != "" {
+		authBJJCredentialStatus = (verifiable.CredentialStatusType)(*authBJJCredentialStatusString)
```
x1m3, line 38: Any value different than "" can be accepted? Or we have a list of allowed values? In that case, we should check only for that allowed values. Otherwise is ok

internal/api_ui/server.go
```diff
@@ -356,6 +356,13 @@ func (s *Server) CreateCredential(ctx context.Context, request CreateCredentialR
 		claimRequestProofs.Iden3SparseMerkleTreeProof = true
 	}
 
+	var credentialStatusType verifiable.CredentialStatusType
+	if request.Body.CredentialStatusType == nil || *request.Body.CredentialStatusType == "" {
+		credentialStatusType = verifiable.Iden3commRevocationStatusV1
```
x1m3, line 361: Same comment as in api/server.go.  Any value for request.Body.CredentialStatusType is acceptable or we have a set of allowed values?

internal/api_ui/server.go
```diff
@@ -703,7 +721,11 @@ func (s *Server) CreateLinkQrCodeCallback(ctx context.Context, request CreateLin
 		return CreateLinkQrCodeCallback400JSONResponse{N400JSONResponse{Message: "error getting reverse hash service settings"}}, nil
 	}
 
-	err = s.linkService.IssueClaim(ctx, request.Params.SessionID.String(), s.cfg.APIUI.IssuerDID, *userDID, request.Params.LinkID, s.cfg.APIUI.ServerURL, rhsSettings.CredentialStatusType)
+	if !s.networkResolver.IsCredentialStatusTypeSupported(rhsSettings, credentialStatusType) {
+		return CreateLinkQrCodeCallback400JSONResponse{N400JSONResponse{Message: fmt.Sprintf("Credential Status '%s' is not supported by the issuer", credentialStatusType)}}, nil
```
x1m3, line 725: log.

internal/api_ui/server.go
```diff
@@ -703,7 +721,11 @@ func (s *Server) CreateLinkQrCodeCallback(ctx context.Context, request CreateLin
...
+	}
+
+	err = s.linkService.IssueClaim(ctx, request.Params.SessionID.String(), s.cfg.APIUI.IssuerDID, *userDID, request.Params.LinkID, s.cfg.APIUI.ServerURL, credentialStatusType)
 	if err != nil {
 		log.Debug(ctx, "error issuing the claim", "error", err)
```
x1m3, line 730: Not for this PR, but this should be a log.Error

x1m3 (commented):
Some comments about validation and logs. That part of code is not well covered with logs. I think it is a good moment to add some of them.

x1m3:
Please, use conventional commits standard as explained in the [contributing doc](https://github.com/0xPolygonID/issuer-node/blob/main/CONTRIBUTING.md)  Just prefixing each PR with feat: chore: fix: etc.. And give a good PR name because this will appear in the changelog of the release.

internal/api/identity.go
volodymyr-basiuk: added logs

internal/api/identity.go
volodymyr-basiuk: check for allowed values added

internal/api_ui/server.go
volodymyr-basiuk: check for allowed values added

internal/api_ui/server.go
volodymyr-basiuk: added

internal/api_ui/server.go
volodymyr-basiuk: fixed

resolvers_settings_sample.yaml
```diff
@@ -58,7 +58,7 @@ polygon:
 #    gasLess: false
 #    transferAmountWei: 1000000000000000000
 #    rhsSettings:
-#      mode: { None | OffChain | OnChain }
+#      mode: { None | OffChain | OnChain | All}
```
martinsaporiti, line 61: when is this value used if the mode is specified when identity/credentials are created?

martinsaporiti (commented):
then the user can create a did issuer using one mode and credentials using another? This works? It was tested?

resolvers_settings_sample.yaml
volodymyr-basiuk: Based on these values we says what mode supported by the issuer. In the `IsCredentialStatusTypeSupported` method we check if param `CredentialStatusType` that was passed by user supported by issuer (for that chain).

ex: issuer supports only OffChain, than if user passes `Iden3OnchainSparseMerkleTreeProof2023` we show 400 error.

volodymyr-basiuk:
> @volodymyr-basiuk then the user can create a did issuer using one mode and credentials using another? This works? It was tested?

Yes, we can create credential in other mode than `authBjjCred` was created (in mode `All`). In such case we publish to both RHS  (look into `BuildPublishers` method, `All` case). I have tested it locally. Also the same logic were in `identity-server`.

vmidyllic (approved):

martinsaporiti:
> > @volodymyr-basiuk then the user can create a did issuer using one mode and credentials using another? This works? It was tested?
> 
> Yes, we can create credential in other mode than `authBjjCred` was created (in mode `All`). In such case we publish to both RHS (look into `BuildPublishers` method, `All` case). I have tested it locally. Also similar logic were in `identity-server`.

Perfect, thanks!

martinsaporiti (approved):
LGTM

PR #706, opened 2024-08-01
feat: add create connection endpoint

volodymyr-basiuk:


martinsaporiti (changes requested):
It is possible to write some tests?

volodymyr-basiuk:
> It is possible to write some tests?

Sure, added

martinsaporiti (approved):
nice, thanks

PR #713, opened 2024-08-06
support multiple contracts for package manager

ilya-korotya:
Temporary pull request


x1m3:
Closed because integration with partner who was needing it is paused.

PR #716, opened 2024-08-07
feat: Implement identities and refactor to support unified APIs 

Alexander-frenki:
https://polygonid.atlassian.net/browse/PID-2297


ui/src/adapters/api/issuers.ts
```diff
@@ -0,0 +1,114 @@
...
+      },
+      method: "GET",
+      signal,
+      url: `${API_VERSION}/identities`,
+    });
+
+    if (response.data === null) {
+      return buildSuccessResponse(getListParser(apiIssuerParser).parse([]));
+    } else {
+      return buildSuccessResponse(getListParser(apiIssuerParser).parse(response.data));
```
volodymyr-basiuk, line 85: `buildSuccessResponse(getListParser(apiIssuerParser).parse(response.data || []));`

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -189,15 +199,17 @@ export function CredentialDetails() {
...
-                    <Button
-                      danger
-                      disabled={revoked}
-                      icon={<IconClose />}
-                      onClick={() => setShowRevokeModal(true)}
-                      type="text"
-                    >
-                      {sm && REVOKE}
-                    </Button>
+                    <Tooltip title={notPuslihedState ? "State not published" : ""}>
```
volodymyr-basiuk, line 202: `"State not published"` move to constants

ui/src/components/issuers.tsx/CreateIssuer.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <>
+      {messageContext}
+
+      <SiderLayoutContent
+        description="Add a new issuer to get the required credential."
+        showBackButton
+        showDivider
+        title={ISSUER_ADD}
+      >
+        <Card className="issuers-card" title="Issuer details">
```
volodymyr-basiuk, line 37: let's move to constants

ui/src/components/issuers.tsx/IssuerForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [showCredentialStatusField, setShowCredentialStatusField] = useState(false);
+
+  const handleFormChange = (changedValues: Partial<IssuerFormData>) => {
+    if (changedValues.type) {
+      if (changedValues.type === IssuerType.BJJ) {
+        setShowCredentialStatusField(true);
+      } else {
+        setShowCredentialStatusField(false);
+      }
+    }
```
volodymyr-basiuk, line 24:
```
if (changedValues?.type === IssuerType.BJJ) {
        setShowCredentialStatusField(true);
      } else {
        setShowCredentialStatusField(false);
      }
```

ui/src/components/issuers.tsx/Issuers.tsx
```diff
@@ -0,0 +1,89 @@
...
+          {isModalOpen && (
+            <Modal
+              centered
+              closable
+              closeIcon={<IconClose />}
+              footer={null}
+              maskClosable
+              onCancel={closeModal}
+              open
+              title="Add new issuer"
```
volodymyr-basiuk, line 80: constant

ui/src/components/issuers.tsx/IssuersTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (parsedIdentifier.success) {
+      handleChange(parsedIdentifier.data);
+    } else {
+      handleChange(null);
+    }
+  };
+
+  useEffect(() => {
+    if (isAsyncTaskDataAvailable(issuersList)) {
+      if (queryParam === null) {
```
volodymyr-basiuk, line 48: `if (!queryParam) {`

ui/src/components/issuers.tsx/IssuersTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+  }, [queryParam, issuersList]);
+
+  const onSearch = useCallback(
+    (query: string) => {
+      setSearchParams((previousParams) => {
+        const previousQuery = previousParams.get(QUERY_SEARCH_PARAM);
+        const params = new URLSearchParams(previousParams);
+
+        if (query === "") {
```
volodymyr-basiuk, line 65: `if (!query) {`

Alexander-frenki: Screenshot 2024-08-07 at 20 04 37

I thought this check made sense with an empty string, so I did it the same way as in other components.

ui/src/components/issuers.tsx/IssuerForm.tsx
Alexander-frenki: changedValues ​​cannot be undefined because this function is called when any field changes

volodymyr-basiuk: then do we need first if ` if (changedValues.type) {` ?

volodymyr-basiuk (approved):

ui/src/adapters/api/credentials.ts
```diff
@@ -87,34 +118,105 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
   signal,
 }: {
   credentialID: string;
   env: Env;
+  identifier: Identifier;
   signal?: AbortSignal;
 }): Promise<Response<Credential>> {
   try {
-    const response = await axios({
+    const response: { data: CredentialDetail } = await axios({
```
amonsosanz, line 130: You can not do this. We always need to assume that the response we get from the API is broken. This is why we create zod parsers and use them to parse the data. To access the data from the credential you need to parse the response from the server before.

ui/src/adapters/api/credentials.ts
```diff
@@ -87,34 +118,105 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+      credentialStatus,
+      credentialSubject,
+      expirationDate,
+      id,
+      issuanceDate,
+      issuer,
+      proof,
+    } = response.data;
+
+    const revocationStatusResponse: { data: RevocationStatus } = await axios({
```
amonsosanz, line 152: Same as [here](https://github.com/0xPolygonID/issuer-node/pull/716/files#r1709003757)

ui/src/contexts/Issuer.tsx
```diff
@@ -0,0 +1,114 @@
...
+type IssuerState = {
+  fetchIssuers: (signal: AbortSignal) => void;
+  handleChange: (identifier: Identifier) => void;
+  identifier: Identifier;
+  issuersList: AsyncTask<Issuer[], AppError>;
+};
+
+const defaultIssuerState: IssuerState = {
+  fetchIssuers: function (): void {},
+  handleChange: function (): void {},
```
amonsosanz, line 28: Would you mind using arrow functions for consistency? `() => void {}`

ui/src/adapters/api/connections.ts
```diff
@@ -52,7 +55,7 @@ export async function getConnection({
       },
       method: "GET",
       signal,
-      url: `${API_VERSION}/connections/${id}`,
+      url: `${API_VERSION}/${identifier}/connections/${id}`,
```
amonsosanz, line 58: How can an `identifier` param used to build a URL be `null`?

I'd remove the type `Identifier` from the domain and use the `string` type everywhere.

Also, I'd rename the param `identifier` as `issuerIdentifier` everywhere. The `identifier` alone does not say much about the param.

ui/src/adapters/api/credentials.ts
```diff
@@ -68,7 +98,8 @@ export const credentialParser = getStrictParser<CredentialInput, Credential>()(
     proofTypes: proofTypeParser,
     refreshService: z
       .object({ id: z.string(), type: z.literal("Iden3RefreshService2023") })
-      .nullable(),
+      .nullable()
+      .default(null),
```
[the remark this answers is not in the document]
Alexander-frenki: in the core API request GET /v1/{identifier}/credentials/{id} does not return the refreshService key if the refresh service has not been configured

ui/src/adapters/api/connections.ts
Alexander-frenki: I did this because there is a case where the identifier is null. When the user does not select an issuer, in this case he will only see the page to select or create an issuer, and before he does so, the identifier will be null

ui/src/adapters/api/credentials.ts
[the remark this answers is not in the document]
Alexander-frenki: but seems like I can remove it

ui/src/adapters/parsers/view.ts
```diff
@@ -425,17 +426,17 @@ export function serializeCredentialIssuance({
       data: {
         credentialSchema,
         credentialSubject: serializedSchemaForm.data === undefined ? {} : serializedSchemaForm.data,
-        expiration: credentialExpiration
-          ? serializeDate(dayjs(credentialExpiration), "date-time")
-          : null,
-        mtProof,
+        expiration: credentialExpiration ? dayjs(credentialExpiration).unix() : null,
```
amonsosanz, line 429: We don't need dayjs to get the timestamp from a date, just do `credentialExpiration.getTime()`

ui/src/assets/icons/building.svg
amonsosanz, line 1: Please respect the original name of the icons in the library so we can always map them back. This file should be named `building-08.svg`

ui/src/components/issuers.tsx/Issuers.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  useEffect(() => {
+    fetchData();
+  }, [fetchData]);
+
+  return (
+    <>
+      {messageContext}
+
+      <SiderLayoutContent description="Description." title={ISSUERS}>
```
amonsosanz, line 67: `description="Description."`?

ui/src/adapters/api/connections.ts
amonsosanz: Ok, we need to think a bit more about this logic and how to properly model it, but we can not end up with a required value such as this identifier that can be `null` app-wide. Do nothing about this for now. I'll propose something when I fully understand the case.

ui/src/adapters/api/credentials.ts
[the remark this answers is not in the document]
amonsosanz: No, if this is as you mentioned, then your change is required.

I've asked @martinsaporiti and apparently, we are no longer honoring a rule we used to have: All the properties in the payloads we move up and down are required but nullable.

If now we have optional properties in the API payloads, then we will need to do what you just did.

ui/src/components/shared/Router.tsx
```diff
@@ -50,15 +62,29 @@ export function Router() {
...
-      </Route>
+          <Route element={<SiderLayout />}>{getLayoutRoutes("sider")}</Route>
+        </>
+      ) : (
+        <>
+          <Route element={<SiderLayout />}>
+            {issuerRoutes.map(({ Component, path }) => (
+              <Route element={<Component />} key={path} path={path} />
+            ))}
+          </Route>
```
amonsosanz, line 83: Instead of creating the `issuerRoutes` list above and then mapping over it here, since this is a very specific and special case I think I would declare the routes here directly to make the code more cohesive and easier to read:

```typescript
<Route element={<SiderLayout />}>
  <Route
    element={<COMPONENTS.issuers />}
    key={ROUTES.issuers.path}
    path={ROUTES.issuers.path}
  />
  <Route
    element={<COMPONENTS.createIssuer />}
    key={ROUTES.createIssuer.path}
    path={ROUTES.createIssuer.path}
  />
</Route>
```

By the way, how is it that we allow navigating to `ROUTES.issuers` when there's no `identifier`? Shouldn't we limit the valid routes to `ROUTES.createIssuer` until `identifier` is available?

ui/src/components/issuers.tsx/CreateIssuer.tsx
amonsosanz, line 1: The parent folder should be named just `issuers` not `issuers.tsx`

ui/src/components/issuers.tsx/IssuersTable.tsx
```diff
@@ -0,0 +1,193 @@
...
+        { blockchain: blockchainA, network: networkA },
+        { blockchain: blockchainB, network: networkB }
+      ) => `${blockchainA}-${networkA}`.localeCompare(`${blockchainB}-${networkB}`),
+      title: "Network",
+    },
+  ];
+
+  const addButton = (
+    <Button icon={<IconPlus />} onClick={handleAddIssuer} type="primary">
+      Add new issuer
```
amonsosanz, line 122: I'm not especially happy having all the literals in constants, but this is something from the past that we should honor until we discard, so this should be `ISSUER_ADD`

ui/src/components/issuers.tsx/Issuers.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { isAsyncTaskDataAvailable } from "src/utils/async";
+import { makeRequestAbortable } from "src/utils/browser";
+import { ISSUERS, ISSUER_ADD } from "src/utils/constants";
+
+export function Issuers() {
+  const env = useEnvContext();
+  const { fetchIssuers, issuersList } = useIssuerContext();
+
+  const [messageAPI, messageContext] = message.useMessage();
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
amonsosanz, line 23: Why do we have a modal and a route to create issuers?

ui/src/components/issuers.tsx/IssuerForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  return (
+    <Form form={form} layout="vertical" onFinish={onSubmit} onValuesChange={handleFormChange}>
+      <Form.Item label="Method" name="method" rules={[{ message: VALUE_REQUIRED, required: true }]}>
+        <Input placeholder="Method" />
```
amonsosanz, line 30: Having in the placeholder the same as we have in the label is useless.

ui/src/components/connections/ConnectionsTable.tsx
```diff
@@ -29,6 +29,7 @@ import { NoResults } from "src/components/shared/NoResults";
 import { SiderLayoutContent } from "src/components/shared/SiderLayoutContent";
 import { TableCard } from "src/components/shared/TableCard";
```
amonsosanz, line 30: We are getting 2 deprecation errors from this component, can you fix them, please?

image

ui/src/adapters/api/connections.ts
Alexander-frenki: I agree with you, will change to empty string instead of null

ui/src/domain/identifier.ts
```diff
@@ -0,0 +1,53 @@
...
+    createdAt: Date;
+    identifier?: string;
+    modifiedAt: Date;
+    previousState?: string;
+    revocationTreeRoot?: string;
+    rootOfRoots?: string;
+    state: string;
+    stateID?: number;
+    status: string;
+    txID?: string;
```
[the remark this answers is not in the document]
Alexander-frenki: actually all this data is not used, it's just a response for the createIssuer function, so the it doesn't matter what this function returns, maybe I can just return the response status?

image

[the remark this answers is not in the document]
amonsosanz: Please just model and parse what we need. If all this is not used, should not be part of our types and parsers.

ui/src/components/issuers.tsx/Issuers.tsx
Alexander-frenki: according to the design, if the user does not have issuer, then a modal window will be shown, if there is, then the creation will lead to the page

ui/src/components/credentials/CredentialDetails.tsx
```diff
@@ -276,7 +294,7 @@ export function CredentialDetails() {
       )}
       {isAsyncTaskDataAvailable(credential) && showRevokeModal && (
         <CredentialRevokeModal
-          credential={credential.data}
+          credential={{ ...credential.data, revoked }}
```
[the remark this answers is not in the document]
Alexander-frenki: in case the status is not published, getRevocationStatus returns false but getCredentials list returns item with revoked: true value, so to display correct view on detail page i set query parameter with correct revoked status and also add tooltip to revoke button which shows not published status.

ui/src/components/shared/Router.tsx
Alexander-frenki: okay i will change this part.

for the `ROUTES.createIssuer` route we have a representation for the case where no issuers are created

ui/src/adapters/parsers/view.ts
Alexander-frenki: but getTime returns timestamp with milliseconds, to request  createCredential we need to provide timestamp with seconds

ui/src/adapters/api/credentials.ts
```diff
@@ -87,34 +118,105 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+        schemaUrl: credentialSchema.id,
+        userID: issuer,
+      })
+    );
+  } catch (error) {
+    return buildErrorResponse(error);
+  }
+}
+
+export async function getRevocationStatus({
```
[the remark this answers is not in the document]
Alexander-frenki: now it used in getCredential request

ui/src/components/issuers.tsx/IssuerForm.tsx
Alexander-frenki: I will request changes from the designer

ui/src/components/issuers.tsx/Issuers.tsx
Alexander-frenki: I will request changes from the designer

ui/src/adapters/api/credentials.ts
```diff
@@ -78,6 +112,46 @@ export const credentialParser = getStrictParser<CredentialInput, Credential>()(
...
+    credentialStatus: z.object({
+      id: z.string(),
+      revocationNonce: z.number(),
+      type: z.nativeEnum(AuthBJJCredentialStatus),
+    }),
+    credentialSubject: z
+      .object({
+        type: z.string(),
+      })
+      .catchall(z.unknown()),
```
amonsosanz, line 135:
```
    credentialSubject: z
      .object({
        type: z.string(),
      })
      .and(z.record(z.unknown())),
```

ui/src/adapters/api/credentials.ts
```diff
@@ -87,12 +161,91 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+    const credentialDetail = await getCredentialDetail({
+      credentialID,
+      env,
+      issuerIdentifier,
+      signal,
+    });
+
+    if (!credentialDetail.success) {
+      return buildErrorResponse(credentialDetail.error);
+    }
```
amonsosanz, line 182:
```
    if (!credentialDetail.success) {
      return credentialDetail;
    }
```

ui/src/adapters/api/credentials.ts
```diff
@@ -87,12 +161,91 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+    const revocationStatus = await getRevocationStatus({
+      env,
+      issuerIdentifier,
+      params: { nonce: credentialStatus.revocationNonce },
+      signal,
+    });
+
+    if (!revocationStatus.success) {
+      return buildErrorResponse(revocationStatus.error);
+    }
```
amonsosanz, line 205: Same as before, no need to rebuild an `ErrorResponse`. When `!revocationStatus.success`, `revocationStatus` is already an `ErrorResponse`.

```
    if (!revocationStatus.success) {
      return revocationStatus;
    }
```

ui/src/adapters/api/credentials.ts
```diff
@@ -87,12 +161,91 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+      credentialParser.parse({
+        createdAt: issuanceDate,
+        credentialSubject,
+        expired,
+        expiresAt: expirationDate,
+        id: id.split(":").at(-1),
+        proofTypes,
+        revNonce: credentialStatus.revocationNonce,
+        revoked,
+        schemaHash: schemaHash.success ? schemaHash.data : "",
```
amonsosanz, line 227: ?

ui/src/adapters/api/credentials.ts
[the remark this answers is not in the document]
amonsosanz: Why nonce is wrapped on a params object?

ui/src/adapters/api/issuers.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+export type CreateIssuer = {
+  authBJJCredentialStatus?: AuthBJJCredentialStatus;
+  blockchain: string;
+  method: string;
+  network: string;
+  type: IssuerType;
+};
```
amonsosanz, line 58: This type does not reflect reality, since `authBJJCredentialStatus` is mandatory, but only when `type` is `BJJ` right?

A better way of modeling this reality would be as follows:

```typescript
export type CreateIssuer = {
  blockchain: string;
  method: string;
  network: string;
} & ({
  authBJJCredentialStatus: AuthBJJCredentialStatus;
  type: IssuerType.BJJ;
} | {
  type: IssuerType.ETH;
});
```

ui/src/adapters/parsers/view.ts
amonsosanz: Ah! Well, we could divide by 1000, but leave it as it is.

ui/src/components/issuers.tsx/Issuers.tsx
amonsosanz: It's ok for now. A completely new design is in the owen.

ui/src/adapters/api/credentials.ts
Alexander-frenki: the getSchemaHash function can return the success: false

amonsosanz: Because it can error. When this happens, we need to propagate the error, not silently swallow it like this.

ui/Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /app
 RUN apk add --update nodejs npm
 RUN apk add --update apache2-utils
 RUN apk add --update libwebp=1.3.2-r0
-RUN apk add --update libx11=1.8.7-r0
+#RUN apk add --update libx11=1.8.7-r0
```
OBrezhniev, line 8: Just remove it completely if it's not needed. Also check if libwebp is needed and remove it completely too or at least remove package version.

OBrezhniev: Also please use `npm ci` instead of `npm install` because it installs exact versions from the lock file and will not suddenly break because some dependency was updated.

ui/src/adapters/api/credentials.ts
```diff
@@ -87,12 +161,95 @@ export const credentialStatusParser = getStrictParser<CredentialStatus>()(
...
+    });
+
+    if (!revocationStatus.success) {
+      return revocationStatus;
+    }
+
+    const proofTypes = proof.map(({ type }) => type);
+
+    const schemaHash = getSchemaHash({
+      id: `${context.at(-1)}#${credentialSubject.type}`,
```
OBrezhniev, line 210: @vmidyllic please review this line, because it looks incorrect to me.

ui/src/adapters/api/credentials.ts
```diff
@@ -426,10 +624,12 @@ const authQRCodeParser = getStrictParser<AuthQRCodeInput, AuthQRCode>()(
 
 export async function createAuthQRCode({
   env,
+  issuerIdentifier,
```
OBrezhniev, line 627: We need to rename "QRCode" endpoints, because they are not generating qr codes. Maybe "QRCode" could be substituted with  "Message". CC: @demonsh

x1m3 (commented):
LGTM.

ui/src/adapters/api/credentials.ts
amonsosanz: Agreed, this should fail for schemas from the schema builder since its ids are a urn:uuid,

`"@id": "urn:uuid:9bbc4cd1-8658-4a48-b013-87603e32e399"`

Ids like this one below are the old ones:

`"@id": "https://raw.githubusercontent.com/anima-protocol/claims-polygonid/main/schemas/json-ld/pol-v1.json-ld#AnimaProofOfLife"`

.env-issuer.sample
amonsosanz, line 1: We should update this ref to polygon id in line 1:

# Visit https://devs.polygonid.com for more configuration information

ui/scripts/deploy.sh
```diff
@@ -5,18 +5,22 @@ ENV_FILENAME="/app/.env"
...
 echo "VITE_BUILD_TAG=$ISSUER_UI_BUILD_TAG" >> $ENV_FILENAME
-echo "VITE_ISSUER_DID=$ISSUER_API_UI_ISSUER_DID" >> $ENV_FILENAME
-echo "VITE_ISSUER_LOGO=$ISSUER_API_UI_ISSUER_LOGO" >> $ENV_FILENAME
-echo "VITE_ISSUER_NAME=$ISSUER_API_UI_ISSUER_NAME" >> $ENV_FILENAME
 echo "VITE_WARNING_MESSAGE=$ISSUER_UI_WARNING_MESSAGE" >> $ENV_FILENAME
 echo "VITE_IPFS_GATEWAY_URL=$ISSUER_UI_IPFS_GATEWAY_URL" >> $ENV_FILENAME
 echo "VITE_SCHEMA_EXPLORER_AND_BUILDER_URL=$ISSUER_UI_SCHEMA_EXPLORER_AND_BUILDER_URL" >> $ENV_FILENAME
 
+# TODO: Remove this envs:
+echo "VITE_ISSUER_DID=VITE_ISSUER_DID" >> $ENV_FILENAME
```
amonsosanz, line 21: Why is this kept?

Alexander-frenki: @martinsaporiti

martinsaporiti: that was kept to avoid errors in the UI until did creation was coded. We can remove it now.

martinsaporiti: and please check `VITE_ISSUER_DID` is not mandatory in `env.ts` file.

.env-issuer.sample
```diff
@@ -13,6 +13,8 @@ ISSUER_LOG_MODE=2
 ISSUER_API_AUTH_USER=user-issuer
 ISSUER_API_AUTH_PASSWORD=password-issuer
 ISSUER_ENVIRONMENT=local
```
amonsosanz, line 15: @martinsaporiti is this used?

ui/src/components/issuers/Onboarding.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useEnvContext } from "src/contexts/Env";
+
+import { useIssuerContext } from "src/contexts/Issuer";
+import { ROUTES } from "src/routes";
+
+import { FINALISE_SETUP } from "src/utils/constants";
+
+const cards = [
+  {
+    icon: <IconIssue />,
```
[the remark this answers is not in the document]
Alexander-frenki: Yes, I agree with you, but it's a global styling, and a lot of colors now look wrong, I discussed  that I would check and refactor all colors, but it will be after I complete this part of the work

that's why this page a did using old colors scheme

ui/src/components/issuers/Onboarding.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  </Flex>
+                </Card>
+                {!isLastCard && lg && <div className="dotted-divider" />}
+              </React.Fragment>
+            );
+          })}
+        </Flex>
+
+        <Divider
+          className="onboarding-divider"
```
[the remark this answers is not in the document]
Alexander-frenki: image

Yes

ui/src/contexts/Issuer.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const IssuerContext = createContext(defaultIssuerState);
+
+export function IssuerProvider(props: PropsWithChildren) {
+  const env = useEnvContext();
+  const [messageAPI, messageContext] = message.useMessage();
+  const [issuersList, setIssuersList] = useState<AsyncTask<Issuer[], AppError>>({
+    status: "pending",
+  });
+  const [issuerIdentifier, setIssuerIdentifier] = useState<IssuerIdentifier>("");
```
amonsosanz, line 47: The selected identity state should be stored in the URL instead of in the context, otherwise, the URLs are not absolute, meaning they do not always point to the same view and data.

ui/src/components/shared/HighlightLink.tsx
```diff
@@ -0,0 +1,56 @@
...
+                icon={<IconLink />}
+                style={{ borderColor: token.colorTextSecondary, color: token.colorTextSecondary }}
+                target="_blank"
+              />
+            )}
+            <Button
+              icon={<IconCopy />}
+              onClick={onCopyToClipboard}
+              style={{ borderColor: token.colorTextSecondary, color: token.colorTextSecondary }}
+            />
```
amonsosanz, line 50: Can we avoid the buttons covering the link text?

ui/src/components/schemas/SchemasTable.tsx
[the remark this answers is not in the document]
Alexander-frenki: https://github.com/0xPolygonID/issuer-node/pull/716/commits/9e8d21385bb2567ea9f415cc69df732e5f630a62
@x1m3 why you did it?

ui/src/components/identities/IdentitiesTable.tsx
```diff
@@ -0,0 +1,205 @@
...
+          menu={{
+            items: [
+              {
+                icon: <IconInfoCircle />,
+                key: "details",
+                label: DETAILS,
+                onClick: () =>
+                  navigate(
+                    generatePath(ROUTES.identityDetails.path, {
+                      issuerID: identifier,
```
amonsosanz, line 125: image

ui/package.json
```diff
@@ -5,8 +5,8 @@
     "ajv": "^8.12.0",
```
amonsosanz, line 5: We have 4 warnings to fix
image

ui/src/components/identities/Onboarding.tsx
```diff
@@ -0,0 +1,156 @@
...
+            size={48}
+            style={{ marginBottom: 16 }}
+          />
+
+          <Typography.Text style={{ fontSize: 30 }}>
+            You successfully installed Issuer Node
+          </Typography.Text>
+          <Typography.Text style={{ fontSize: 20 }} type="secondary">
+            Here&apos;s what you&apos;re going to be able to do with the issuer node, once you
+            finalize your setup
```
amonsosanz, line 77: > Here’s what you’ll be able to do with the issuer node once you finalize your setup

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,138 @@
-import { Avatar, Row, Space, Typography } from "antd";
+import { Avatar, Card, Dropdown, Flex, Tag, Tooltip, Typography, theme } from "antd";
```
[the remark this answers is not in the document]
Alexander-frenki: I don't know why but looks good on my laptop and [on testing env](https://issuer-node-testing.privado.id/)
image

ui/src/domain/credential.ts
```diff
@@ -1,15 +1,26 @@
...
 
+export type Proof = {
+  type: CredentialProofType;
+};
+
 export type Credential = {
   createdAt: Date;
-  credentialSubject: Record<string, unknown>;
+  credentialSubject: {
+    type: string;
```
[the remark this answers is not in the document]
Alexander-frenki: credentialParser

image

ui/src/domain/identity.ts
```diff
@@ -0,0 +1,38 @@
+export type Identifier = string;
```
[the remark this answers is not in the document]
Alexander-frenki: sorry, but I can't understand what the problem is

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,138 @@
...
 
 export function UserDisplay() {
   const { issuer } = useEnvContext();
+  const { handleChange, identifier, identitiesList, identityDisplayName } = useIdentityContext();
+  const navigate = useNavigate();
+  const { token } = theme.useToken();
+
+  const identityItems = isAsyncTaskDataAvailable(identitiesList)
+    ? identitiesList.data
+        .toSorted((item) => (item.identifier === identifier ? -1 : 0))
```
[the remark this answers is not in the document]
Alexander-frenki: When I did this I thought it might be useful when you have a lot of personalities, but if you think it's redundant I can remove it.

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,138 @@
...
+    {
+      key: "test",
+      label: (
+        <Flex gap={16}>
+          <IconPlus style={{ height: 20, width: 20 }} />
+          <Typography.Text>{IDENTITY_ADD}</Typography.Text>
+        </Flex>
+      ),
+      onClick: () => navigate(ROUTES.createIdentity.path),
+    },
```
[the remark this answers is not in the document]
Alexander-frenki: this btn
Screenshot 2024-10-09 at 21 20 19

ui/src/components/shared/Router.tsx
```diff
@@ -50,15 +59,27 @@ export function Router() {
...
 
-      <Route element={<FullWidthLayout background="bg-light" />}>
-        {getLayoutRoutes("fullWidthGrey")}
-      </Route>
+          <Route element={<SiderLayout />}>{getLayoutRoutes("sider")}</Route>
+        </>
+      ) : (
+        <>
+          <Route element={<FullWidthLayout background="background-grey" />}>
+            <Route element={<COMPONENTS.onboarding />} path={ROUTES.onboarding.path} />
```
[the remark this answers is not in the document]
Alexander-frenki: I think we need a page because we also need to render layout

ui/scripts/deploy.sh
amonsosanz: And the `did` prop from the `issuer` in the `Env` type.

ui/src/components/issuers/Onboarding.tsx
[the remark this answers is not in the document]
amonsosanz: 🤯 I never used sass to build a className like `.onboarding-divider` nesting 2 sass rules.

I mean, I used sass nesting, but always prepending the inner name with a dot, so it applies to descendants. Never used a dash to compile a flat className. Thx!

I'm a sass newbie!

ui/src/components/schemas/SchemasTable.tsx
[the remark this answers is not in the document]
amonsosanz: I'm curios too @x1m3!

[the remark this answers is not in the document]
x1m3: I saw that npm run build was complaining about it. 

Maybe it is wrong change? Please, revert.

ui/package.json
```diff
@@ -5,8 +5,8 @@
...
     "ajv-formats": "^2.1.1",
     "ajv-formats-draft2019": "^1.6.1",
-    "antd": "^5.18.0",
-    "axios": "^1.6.1",
+    "antd": "^5.20.4",
+    "axios": "^1.7.4",
```
amonsosanz, line 9: We should update all libraries, at least to their newest compatible version, before the release.

ui/src/components/schemas/SchemasTable.tsx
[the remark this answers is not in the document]
amonsosanz: Yep, it was deprecated and produced the warning. Already reverted.

ui/src/components/shared/UserDisplay.tsx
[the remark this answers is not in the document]
amonsosanz: Interesting. What browser are you using? I use:

> Chrome 129.0.6668.90 (Official Build) (arm64)

ui/src/domain/credential.ts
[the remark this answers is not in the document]
amonsosanz: Ok, I understand, but we should not rely on this type, but on the type property in the root of VC:

`[“VerifiableCredential”, “type”]`

ui/src/domain/identity.ts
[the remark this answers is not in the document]
amonsosanz: We don't need a type `Identifier` that equals `string`.

This adds unnecessary cognitive load since I need to go to the type `Identifier` to realize it is a `string`.

Wherever we use `Identifier` we should use `string` instead.

ui/src/adapters/api/identities.ts
```diff
@@ -0,0 +1,193 @@
...
+    return buildSuccessResponse(undefined);
+  } catch (error) {
+    return buildErrorResponse(error);
+  }
+}
+
+export const supportedNetworkParser = getStrictParser<SupportedNetwork>()(
+  z.object({
+    blockchain: z.string(),
+    networks: z.array(z.string()).nonempty(),
```
amonsosanz, line 167: This parser better represents the type:

`z.tuple([z.string()]).rest(z.string())`

ui/src/components/shared/UserDisplay.tsx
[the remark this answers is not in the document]
amonsosanz: Yes please, remove it. It's confusing that the list gets resorted whenever you select an identity.

ui/src/components/shared/UserDisplay.tsx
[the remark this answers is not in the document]
amonsosanz: Ah! These are the scrollbars that Safari hides.

Ok, at least I think we can safely remove the horizontal one.

ui/src/adapters/api/credentials.ts
```diff
@@ -52,30 +52,92 @@ const proofTypeParser = getStrictParser<ProofTypeInput[], ProofType[]>()(
...
+          id,
+          proofTypes,
+          refreshService: refreshService,
+          revNonce: credentialStatus.revocationNonce,
+          revoked,
+          schemaHash,
+          schemaType: credentialSubject.type,
+          schemaUrl: credentialSchema.id,
+          userID: issuer,
+        };
```
amonsosanz, line 138: Now that the API has changed that drastically, maybe we can consider updating our domain type, so our structure better matches what the API provides.

For example, mapping `createdAt` to `issuanceDate` does not make much sense, since the previous value was a database record creation date while the current value represents when the credential was issued.

I would create a parser for the vc object that processes the types that require it, like dates, and use it on the wrapper type that would better match the current API structure.

ui/src/adapters/api/credentials.ts
```diff
@@ -133,15 +199,15 @@ export async function getCredentials({
       },
       method: "GET",
       params: new URLSearchParams({
-        ...(did !== undefined ? { did } : {}),
+        ...(credentialSubject !== undefined ? { credentialSubject } : {}),
```
amonsosanz, line 202: This is wrong. @martinsaporiti @x1m3 can you please clarify why this param has been renamed?

amonsosanz: [Slack discussion on the topic](https://privadoid.slack.com/archives/C074GNHEN82/p1728633454050889)

ui/src/adapters/api/credentials.ts
```diff
@@ -447,43 +532,54 @@ export async function createAuthQRCode({
...
 const issuedQRCodeParser = getStrictParser<IssuedQRCodeInput, IssuedQRCode>()(
   z
     .object({
-      qrCodeLink: z.string(),
       schemaType: z.string(),
+      universalLink: z.string(),
     })
-    .transform(({ qrCodeLink, schemaType }) => ({ qrCode: qrCodeLink, schemaType: schemaType }))
+    .transform(({ schemaType, universalLink }) => ({
+      qrCode: universalLink,
```
amonsosanz, line 546: Instead of doing this transformation, we should rename our domain entity so it also names this `universalLink`

ui/src/adapters/api/identities.ts
```diff
@@ -0,0 +1,193 @@
...
+    blockchain: z.string(),
+    credentialStatusType: z.nativeEnum(CredentialStatusType),
+    displayName: z.string(),
+    identifier: z.string(),
+    method: z.nativeEnum(Method),
+    network: z.string(),
+  })
+);
+
+export const identifierParser = getStrictParser<Identifier>()(z.string());
```
amonsosanz, line 32: To be removed

ui/src/adapters/api/identities.ts
```diff
@@ -0,0 +1,193 @@
...
+      baseURL: env.api.url,
+      headers: {
+        Authorization: buildAuthorizationHeader(env),
+      },
+      method: "GET",
+      signal,
+      url: `${API_VERSION}/identities`,
+    });
+
+    return buildSuccessResponse(getListParser(identityParser).parse(response.data || []));
```
amonsosanz, line 52: Why doing this `|| []`?

ui/src/adapters/api/identities.ts
```diff
@@ -0,0 +1,193 @@
...
+export type CreateIdentity = {
+  blockchain: string;
+  credentialStatusType: CredentialStatusType;
+  displayName: string;
+  method: string;
+  network: string;
+  type: IdentityType;
+};
+
+type CreatedIdentityResponse = {
```
amonsosanz, line 67: `CreateIdentityResponse`

ui/src/adapters/api/identities.ts
Alexander-frenki: because the request returns null when identities do not exist

ui/scripts/deploy.sh
amonsosanz: This is pending from this comment:

> We also need to update `.env-ui.sample`, most data there is outdated.
> Same with `ui/.env.sample` which is empty.

ui/src/components/identities/IdentitiesTable.tsx
```diff
@@ -0,0 +1,205 @@
...
+            Add a new issuer to get the required credential
+          </Typography.Text>
+
+          {addButton}
+        </>
+      }
+      isLoading={isAsyncTaskStarting(identitiesList)}
+      onSearch={onSearch}
+      query={queryParam}
+      searchPlaceholder="Search Issuer"
```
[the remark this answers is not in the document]
amonsosanz: Let me insist,

`Search identity name, DID, etc...`

ui/src/adapters/api/credentials.ts
```diff
@@ -30,19 +31,17 @@ import {
...
-
-const proofTypeParser = getStrictParser<ProofTypeInput[], ProofType[]>()(
+const proofTypeParser = getStrictParser<CredentialProofType[], ProofType[]>()(
   z
-    .array(z.union([z.literal("BJJSignature2021"), z.literal("SparseMerkleTreeProof")]))
+    .array(z.nativeEnum(CredentialProofType))
     .min(1)
     .transform((values) =>
       values.map((value) => {
         switch (value) {
```
demonsh, line 40: we should always display the correct type of Proof. They should not be replaced with SIG/MTP. 
It should be the proper full type `BJJSignature2021`, `Iden3SparseMerkleTreeProof` or other type of proof form the credential `proof` section

ui/src/adapters/api/credentials.ts
amonsosanz: Maybe I did not explain myself well. I would create a separate type and parser for the `vc` object and then add a `vc` property to our domain Credential type, so we don't need to do all this flattening.

ui/src/adapters/api/credentials.ts
amonsosanz: As decided in the the thread, we leave it as it's.

ui/src/adapters/api/credentials.ts
amonsosanz: Now we can remove the type `IssuedQRCodeInput`

ui/src/adapters/api/credentials.ts
amonsosanz: Let's remove the type `ProofType` and use the enum `CredentialProofType` from the SDK everywhere, as we do in the web wallet.

amonsosanz: I just checked and no, we don't even have the JS SDK as a dependency on the schema builder yet. Let's leave it as you did it for now, but this needs to be replaced by the SDK enum along with several more changes.

amonsosanz: We should replace these references too:

image

_________

image

Alexander-frenki: image
it's already renamed in credential link details, or do you mean that we need to remove (SIG) and (MTP) in form?

ui/src/adapters/api/credentials.ts
Alexander-frenki: can we make this refactoring as a tech debt in the future PR to finish this one?

ui/package.json
Alexander-frenki: updated but without eslint and vite-tsconfig-paths packages because that packages make an issues, as I understand @iden3/eslint-config-react-ts should be updated for using a new eslint version. vite-tsconfig-paths also have an issue, and information that I found that fsevents that rollup used should be updated, but it's not our direct dependence

amonsosanz: What issues did you encounter?

ui/src/adapters/api/credentials.ts
amonsosanz: Ok, please ping me when the PR is ready. I'll leave this comment open here.

ui/src/adapters/api/credentials.ts
amonsosanz: Yes, everywhere. Consistency is crucial.

ui/src/components/credentials/CredentialIssuedQR.tsx
```diff
@@ -6,12 +6,14 @@ import { CredentialQR } from "src/components/credentials/CredentialQR";
 import { ErrorResult } from "src/components/shared/ErrorResult";
```
amonsosanz, line 6: I think this component should be removed. Let's wait for [the conclusions of this thread](https://privadoid.slack.com/archives/C074GNHEN82/p1728905716442129).

ui/src/components/credentials/CredentialsTable.tsx
```diff
@@ -171,7 +174,12 @@ export function CredentialsTable() {
                 key: "details",
                 label: DETAILS,
                 onClick: () =>
-                  navigate(generatePath(ROUTES.credentialDetails.path, { credentialID: id })),
+                  navigate({
+                    pathname: generatePath(ROUTES.credentialDetails.path, {
+                      credentialID: id,
+                    }),
+                    search: new URLSearchParams({ revoked: `${credential.revoked}` }).toString(),
```
amonsosanz, line 181: I guess this is a leftover

ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -450,16 +460,16 @@ export function IssueCredentialForm({
                         >
                           <Checkbox.Group>
                             <Space direction="vertical">
-                              <Checkbox value="SIG">
-                                <Typography.Text>Signature-based (SIG)</Typography.Text>
+                              <Checkbox value={CredentialProofType.BJJSignature2021}>
+                                <Typography.Text>Signature-based</Typography.Text>
```
amonsosanz, line 464: Shouldn't we show here the real proof type @demonsh?

ui/src/components/identities/IdentitiesTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+  );
+
+  const tableColumns: TableColumnsType<Identity> = [
+    {
+      dataIndex: "displayName",
+      key: "displayName",
+      render: (displayName: Identity["displayName"]) => (
+        <Typography.Text strong>{displayName}</Typography.Text>
+      ),
+      sorter: ({ displayName: a }, { displayName: b }) => (a || "").localeCompare(b || ""),
```
amonsosanz, line 84: Always fallback displayName to the shortened DID

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,112 @@
...
+              gap={16}
+              justify="space-between"
+            >
+              <Flex
+                style={{
+                  width: 242,
+                }}
+                vertical
+              >
+                <Tooltip title={identity.displayName}>
```
amonsosanz, line 37: Avoid tooltips that show the same underlying data

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,112 @@
...
+                  <Typography.Text ellipsis>{identity.displayName}</Typography.Text>
```
amonsosanz, line 38: What if there's no displayName? Show shortened DID instead.

ui/src/components/shared/UserDisplay.tsx
```diff
@@ -1,18 +1,112 @@
...
+            <Avatar shape="square" size="large" src={issuer.logo} />
+            <IconChevron />
+          </Flex>
+          <Flex gap={4} vertical>
+            <Typography.Text
+              ellipsis={{ tooltip: identityDisplayName }}
+              style={{ fontWeight: 600 }}
+              type="secondary"
+            >
+              {identityDisplayName}
```
amonsosanz, line 82: What if there's no displayName? Show shortened DID instead. This is not very useful:
image

ui/src/components/credentials/IssueCredential.tsx
```diff
@@ -287,7 +301,14 @@ export function IssueCredential() {
             }
 
             case "summary": {
-              return isAsyncTaskDataAvailable(linkID) && <Summary linkID={linkID.data} />;
+              return (
+                isAsyncTaskDataAvailable(authQRCode) && (
+                  <Summary
+                    deepLink={authQRCode.data.deepLink}
+                    universalLink={authQRCode.data.universalLink}
```
amonsosanz, line 308: We could be passing just `AuthQRCode` right?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { AppError, Identity } from "src/domain";
+import { ROUTES } from "src/routes";
+import { AsyncTask, isAsyncTaskDataAvailable } from "src/utils/async";
+import { isAbortedError, makeRequestAbortable } from "src/utils/browser";
+import { IDENTIFIER_SEARCH_PARAM, ROOT_PATH } from "src/utils/constants";
+import { buildAppError } from "src/utils/error";
+
+type IdentityState = {
+  fetchIdentities: (signal: AbortSignal) => void;
+  handleChange: (identifier: string) => void;
```
amonsosanz, line 24: This is selecting an identity, right? Can we better name it? Maybe `selectIdentity`?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,157 @@
...
+  identifier: string;
+  identitiesList: AsyncTask<Identity[], AppError>;
+  identityDisplayName: string;
```
amonsosanz, line 27: Why do we need this property if it's part of the selected identity?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { AsyncTask, isAsyncTaskDataAvailable } from "src/utils/async";
+import { isAbortedError, makeRequestAbortable } from "src/utils/browser";
+import { IDENTIFIER_SEARCH_PARAM, ROOT_PATH } from "src/utils/constants";
+import { buildAppError } from "src/utils/error";
+
+type IdentityState = {
+  fetchIdentities: (signal: AbortSignal) => void;
+  handleChange: (identifier: string) => void;
+  identifier: string;
+  identitiesList: AsyncTask<Identity[], AppError>;
```
amonsosanz, line 26: `identityList` is the right spelling in english.

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const IdentityContext = createContext(defaultIdentityState);
+
+export function IdentityProvider(props: PropsWithChildren) {
+  const env = useEnvContext();
+  const [messageAPI, messageContext] = message.useMessage();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const [identitiesList, setIdentitiesList] = useState<AsyncTask<Identity[], AppError>>({
```
amonsosanz, line 45: Same as [here](https://github.com/0xPolygonID/issuer-node/pull/716/files#r1799591695)

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,157 @@
...
+    status: "pending",
+  });
+  const [identifier, setIdentifier] = useState("");
```
amonsosanz, line 48: Why do we need to declare this state if we are storing the selected identifier in the URL?

ui/src/domain/credential.ts
```diff
@@ -1,19 +1,22 @@
 export type CredentialsTabIDs = "issued" | "links";
 
-export type ProofType = "MTP" | "SIG";
+export enum CredentialProofType {
```
amonsosanz, line 3: Let's name this `ProofType` as in the SDK

ui/src/components/identities/IdentitiesTable.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const queryParam = searchParams.get(QUERY_SEARCH_PARAM);
+
+  useEffect(() => {
+    if (isAsyncTaskDataAvailable(identitiesList)) {
+      if (!queryParam) {
+        setFilteredIdentifiers(identitiesList.data);
+      } else {
+        const filteredData = identitiesList.data.filter((issuer: Identity) =>
+          Object.values(issuer).some((value: string | null) => value?.includes(queryParam))
```
amonsosanz, line 52: We should support case insensitive search

ui/src/contexts/Identity.tsx
Alexander-frenki: because when we navigate to some page we lose the identifier in the URL and set it each time from a state

ui/src/contexts/Identity.tsx
Alexander-frenki: we don't have selected identity, we have selected identifier instead

ui/src/contexts/Identity.tsx
Alexander-frenki: but we keep there a list and few function, I am not sure that fetchIdentities connected to selectIdentity name

ui/Dockerfile
OBrezhniev: @x1m3 Still libwebp is there with strict version (which could already be removed from the package repos). And npm install is also there.

ui/src/components/credentials/IssueCredentialForm.tsx
demonsh: Yes. Please show the real proof type

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
     return buildErrorResponse(error);
   }
 }
 
-type AuthQRCodeInput = Omit<AuthQRCode, "linkDetail"> & {
-  linkDetail: { proofTypes: ProofTypeInput[]; schemaType: string };
-};
-
-export type AuthQRCode = {
+export type AuthMessage = {
```
OBrezhniev, line 468: Rename to `AuthRequestMessage`

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
+  deepLink: string;
   linkDetail: { proofTypes: ProofType[]; schemaType: string };
-  qrCodeLink: string;
   qrCodeRaw: string;
```
OBrezhniev, line 471: There's no QR Codes, rename to `message`

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
+const authMessageParser = getStrictParser<AuthMessage>()(
   z.object({
-    linkDetail: z.object({ proofTypes: proofTypeParser, schemaType: z.string() }),
-    qrCodeLink: z.string(),
+    deepLink: z.string(),
+    linkDetail: z.object({
+      proofTypes: z.array(z.nativeEnum(ProofType)),
+      schemaType: z.string(),
+    }),
     qrCodeRaw: z.string(),
```
OBrezhniev, line 482: Rename

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
+  deepLink: string;
   linkDetail: { proofTypes: ProofType[]; schemaType: string };
-  qrCodeLink: string;
   qrCodeRaw: string;
-  sessionID: string;
+  universalLink: string;
 };
 
-const authQRCodeParser = getStrictParser<AuthQRCodeInput, AuthQRCode>()(
+const authMessageParser = getStrictParser<AuthMessage>()(
```
OBrezhniev, line 475: authRequestMessageParser

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
+      schemaType: z.string(),
+    }),
     qrCodeRaw: z.string(),
-    sessionID: z.string(),
+    universalLink: z.string(),
   })
 );
 
-export async function createAuthQRCode({
+export async function createAuthMessage({
```
OBrezhniev, line 487: createAuthRequestMessage

ui/src/adapters/api/credentials.ts
```diff
@@ -396,138 +457,103 @@ export async function createLink({
...
       axios({
         baseURL: env.api.url,
+        headers: {
+          Authorization: buildAuthorizationHeader(env),
+        },
         method: "GET",
-        params: { type: "link" },
         signal,
-        url: `${API_VERSION}/credentials/${credentialID}/qrcode`,
+        url: `${API_VERSION}/identities/${identifier}/credentials/${credentialID}/qrcode`,
```
OBrezhniev, line 538: @x1m3 Are we still keeping "qrcode" in the url? We should rename it to something else.

OBrezhniev: maybe `.../fetchMessage` and also add the message itself there? WDYT @demonsh?

ui/src/components/credentials/IssueCredentialForm.tsx
OBrezhniev: we should

OBrezhniev (changes requested):

ui/src/components/credentials/IssueCredentialForm.tsx
amonsosanz: What names do you mean by proper names? `BJJSignature2021` and `Iden3SparseMerkleTreeProof` instead of `SIG` and `MTP`?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [identityList, setIdentityList] = useState<AsyncTask<Identity[], AppError>>({
+    status: "pending",
+  });
+  const [identifier, setIdentifier] = useState("");
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const identity =
+    (identityList.status === "successful" || identityList.status === "reloading") &&
+    identityList.data.find((identity) => identity.identifier === identifier);
+  const identityDisplayName = identity && identity.displayName ? identity.displayName : "";
```
amonsosanz, line 63: Again, we're failing to properly fallback the displayName. Let's change this because it's obviously problematic.

Instead of falling back to the short did everywhere, let's do it once, in the adapter, so the view does not need to worry about this.

This means the displayName is required in our domain entity, and we do the fallback only once and forget about it everywhere else.

ui/src/contexts/Identity.tsx
amonsosanz: Storing two properties that can potentially be out of sync it's not a good idea IMO. I see 2 options here:

  1. We store a full "selected identity" in the state, so anyone can use it directly.
  2. We store only the identifier and have a helper function that provided an identifier returns the selected identity.

I prefer 1.

ui/src/contexts/Identity.tsx
Alexander-frenki: we don't need to fall back to short did here, because it does not make sense to show a short did twice

ui/src/contexts/Identity.tsx
amonsosanz: Agreed on a meet.

ui/src/contexts/Identity.tsx
amonsosanz: Agreed on a meet.

ui/src/components/issuer-state/IssuerState.tsx
```diff
@@ -72,37 +108,83 @@ export function IssuerState() {
...
     {
       dataIndex: "txID",
-      ellipsis: { showTitle: false },
+      ellipsis: { showTitle: true },
       key: "txID",
       render: (txID: Transaction["txID"]) => (
-        <Typography.Text strong>
+        <>
           {txID ? (
-            <Link target="_blank" to={`${env.blockExplorerUrl}/tx/${txID}`}>
```
amonsosanz, line 99: After we removed this, we need to remove the `blockExplorerUrl` from the env.

ui/src/components/credentials/IssueCredentialForm.tsx
OBrezhniev: yes, BJJSignature2021 and Iden3SparseMerkleTreeProof

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+  hasAsyncTaskFailed,
+  isAsyncTaskDataAvailable,
+  isAsyncTaskStarting,
+} from "src/utils/async";
+import { isAbortedError, makeRequestAbortable } from "src/utils/browser";
+import { VALUE_REQUIRED } from "src/utils/constants";
+
+const initialValues: IdentityFormData = {
+  blockchain: "",
+  credentialStatusType: CredentialStatusType.Iden3OnchainSparseMerkleTreeProof2023,
```
amonsosanz, line 20: Since apparently we're not fixing [this](https://privadoid.slack.com/archives/C07AFTCMYGJ/p1727863701898819), let's at least default to `CredentialStatusType["Iden3commRevocationStatusV1.0"]` as the backend does.

ui/src/components/schemas/ImportSchemaPreview.tsx
```diff
@@ -29,7 +29,9 @@ export function ImportSchemaPreview({
   const bigintResult = getBigint(jsonLdType);
   const bigint = bigintResult && bigintResult.success ? bigintResult.data : null;
   const schemaHashResult = getSchemaHash(jsonLdType);
-  const schemaHash = schemaHashResult && schemaHashResult.success ? schemaHashResult.data : null;
+  const schemaHash = schemaHashResult.success
+    ? schemaHashResult.data
+    : schemaHashResult.error.message;
```
amonsosanz, line 34: This is no place for an error. Since this error is not a blocker, we better show a toast and default to `"-"`

ui/src/components/shared/CredentialRevokeModal.tsx
```diff
@@ -18,22 +20,28 @@ export function CredentialRevokeModal({
...
 
-    void revokeCredential({ env, nonce }).then((response) => {
+    void revokeCredential({ env, identifier, nonce }).then((response) => {
       if (response.success) {
         onClose();
         onRevoke();
-
+        setSearchParams((previousParams) => {
+          const params = new URLSearchParams(previousParams);
+          params.set(REVOKED_SEARCH_PARAM, "true");
```
amonsosanz, line 42: Another leftover?

Please make sure you remove everything about this, including `REVOKED_SEARCH_PARAM`.

ui/src/components/shared/SiderMenu.tsx
```diff
@@ -93,24 +110,24 @@ export function SiderMenu({
...
       >
         <Col>
           <UserDisplay />
 
-          <Divider />
-
           <Menu
             items={[
               {
+                disabled: !identifier,
```
amonsosanz, line 122: What's this for?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,191 @@
...
```
amonsosanz, line 65: Why is this still here? Didn't we agree to remove it?

ui/src/contexts/Identity.tsx
```diff
@@ -0,0 +1,191 @@
...
+    identityList,
+    selectIdentity,
+    fetchIdentities,
+    getSelectedIdentity,
+  ]);
+
+  return (
+    <>
+      {messageContext}
+      {(identityList.status === "successful" || identityList.status === "reloading") && (
```
amonsosanz, line 182: Why is this check required?

x1m3 (approved):

PR #719, opened 2024-08-08
Add api test

Lechu13:


martinsaporiti (approved):

x1m3:
Please: Rename it using conventional commits like other PRs. This way, this feature will be automatically documented on changelog.

x1m3:
In some way, I see this tests as duplicated from the existing ones in the backend go code. This tests are also end to end tests that are launching a server and doing http requests to it.

What it the extra value? When implementing a new feature, are we going to create an end to end test in go code but also in playwright?

PR #762, opened 2024-08-31
fix: ui dependency

Tickether:


x1m3:
@Tickether Thanks for your contribution. We are about to do a new release and this change was already included on it.

PR #788, opened 2024-09-17
Add docker compose with privadoid registry images

martinsaporiti:


infrastructure/local/docker-compose-full.yml
```diff
@@ -0,0 +1,107 @@
...
+  postgres:
+    image: postgres:14-alpine
+    hostname: postgres
+    environment:
+      - PGPORT=5432
+      - PGUSER=postgres
+      - POSTGRES_HOST_AUTH_METHOD=trust
+      - POSTGRES_USER=postgres
+    ports:
+      - "5432:5432"
```
OBrezhniev, line 11: please remove port forwarding, this is a common issue with the issuer node setup that users have another postgres already running. Instead we can "expose" ports to other containers in the project: https://docs.docker.com/reference/compose-file/services/#expose

infrastructure/local/docker-compose-full.yml
```diff
@@ -0,0 +1,107 @@
...
+    healthcheck:
+      test: [ "CMD-SHELL", "pg_isready" ]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+
+  redis:
+    image: redis:6-alpine
+    ports:
+      - "6379:6379"
```
OBrezhniev, line 24: same here

infrastructure/local/docker-compose-full.yml
```diff
@@ -0,0 +1,107 @@
...
+      timeout: 5s
+      retries: 5
+    depends_on:
+      - api
+
+  pending_publisher:
+    image: privadoid/issuernode
+    pull_policy: always
+    ports:
+      - "3005:3005"
```
OBrezhniev, line 88: probably port doesn't need to be exposed

infrastructure/local/docker-compose-full.yml
```diff
@@ -0,0 +1,107 @@
...
+    command: sh -c "./migrate && ./platform"
+    depends_on:
+      - postgres
+      - redis
+
+  notifications:
+    image: privadoid/issuernode
+    pull_policy: always
+    ports:
+      - "3004:3004"
```
OBrezhniev, line 70: Do we need to expose them?

martinsaporiti: we don't

PR #828, opened 2024-10-21
chore: resolvers settings file and update core lib

martinsaporiti:


go.mod
```diff
@@ -24,8 +24,8 @@ require (
 	github.com/iden3/contracts-abi/state/go/abi v1.0.2-0.20231006071010-f511d08ca36f
 	github.com/iden3/go-circuits/v2 v2.3.0
 	github.com/iden3/go-iden3-auth/v2 v2.4.1
-	github.com/iden3/go-iden3-core/v2 v2.2.0
-	github.com/iden3/go-iden3-crypto v0.0.16
+	github.com/iden3/go-iden3-core/v2 v2.3.1
```
OBrezhniev, line 27: please update other iden3 deps too

resolvers_settings_sample.yaml
```diff
@@ -31,14 +71,119 @@ polygon:
     rpcResponseTimeout: 5s
     waitReceiptCycleTime: 30s
     waitBlockCycleTime: 30s
-    gasLess: false
+    gasLess: true
```
OBrezhniev, line 74: it shouldn't be gasLess

resolvers_settings_sample.yaml
```diff
@@ -14,11 +14,51 @@ polygon:
     gasLess: true
     rhsSettings:
       mode: None
-      contractAddress: 0x16A1ae4c460C0a42f0a87e69c526c61599B28BC9
+      contractAddress: 0x3d3763eC0a50CE1AdF83d0b5D99FBE0e3fEB43fb
```
OBrezhniev, line 17: On all networks rhs should be 0x7dF7...

OBrezhniev (changes requested):

resolvers_settings_sample.yaml
OBrezhniev: Also I don't remember if we have new rhs url for offchain.

martinsaporiti: I think we dont have a new domain for rhs.

resolvers_settings_sample.yaml
```diff
@@ -34,11 +74,116 @@ polygon:
...
+    defaultGasLimit: 600000
+    confirmationTimeout: 10s
+    confirmationBlockCount: 5
+    receiptTimeout: 600s
+    minGasPrice: 0
+    maxGasPrice: 1000000
+    rpcResponseTimeout: 5s
+    waitReceiptCycleTime: 30s
+    waitBlockCycleTime: 30s
+    gasLess: true
```
OBrezhniev, line 136: false

resolvers_settings_sample.yaml
```diff
@@ -34,11 +74,116 @@ polygon:
...
```
OBrezhniev, line 95: gasLess: false

resolvers_settings_sample.yaml
```diff
@@ -34,11 +74,116 @@ polygon:
...
```
OBrezhniev, line 177: false

OBrezhniev (changes requested):

OBrezhniev (approved):

PR #836, opened 2024-10-28
feat: Accepted rhs modes in supported networks endpoint

x1m3:


martinsaporiti (approved):
LGTM

x1m3 (commented):
UI changes tested. LGTM

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -82,13 +114,20 @@ export function IdentityForm({
...
   }, [fetchNetworks]);
 
-  if (hasAsyncTaskFailed(supportedNetworks)) {
+  if (
+    hasAsyncTaskFailed(supportedNetworks) ||
+    (isAsyncTaskDataAvailable(supportedNetworks) && !supportedNetworks.data.length)
+  ) {
+    const errorMessage = hasAsyncTaskFailed(supportedNetworks)
+      ? supportedNetworks.error.message
+      : "";
```
amonsosanz, line 123: An error msg like `""` is illegal. We must model this better.

amonsosanz: I'd go this way:

```typescript
switch (supportedNetworks.status) {
  case "pending":
  case "loading": {
    return (
      <Card className="centered">
        <LoadingResult />
      </Card>
    );
  }
  case "failed": {
    return (
      <Card className="centered">
        <ErrorResult
          error={[
            "An error occurred while downloading the supported networks from the API:",
            supportedNetworks.error.message,
          ].join("\n")}
        />
      </Card>
    );
  }
  case "reloading":
  case "successful": {
    if (supportedNetworks.data.length === 0) {
      return (
        <Card className="centered">
          <ErrorResult error="No supported networks found" />
        </Card>
      );
    }
   ....
  }
```

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -82,13 +114,20 @@ export function IdentityForm({
...
+
     return (
       <Card className="centered">
```
amonsosanz, line 126: We don't want/need these double Card wrappers:

image

image

Let's return just the `ErrorResult` and `LoadingResult` components.

ui/src/components/identities/IdentityForm.tsx
amonsosanz: In reality, checking `supportedNetworks.data.length` is not even a job of the view, but of the parser.

Since we're already doing that check in the type `[Network, ...Network[]]` and therefore in the parser, this check is not necessary here. Whenever we get a 200 and the list is empty we will get a zod error in the view.

Please, keep the switch/case statement for clarity.

ui/src/domain/identity.ts
```diff
@@ -3,33 +3,32 @@ export enum IdentityType {
...
 
+export type Network = {
+  name: string;
+  rhsMode: [string, ...string[]];
+};
+
 export type SupportedNetwork = {
   blockchain: string;
-  networks: [string, ...string[]];
+  networks: [Network, ...Network[]];
```
amonsosanz, line 18: Having the type `SupportedNetwork` that has the property `networks` is very confusing.

Let's rename the `SupportedNetwork` type as `Blockchain`, so we have a blockchain entity (Polygon) that has networks on it (Mainet, Amoy, etc).

Please make sure you propagate these renames to all the parsers, constants, function parameters, etc.

ui/src/domain/identity.ts
```diff
@@ -3,33 +3,32 @@ export enum IdentityType {
   ETH = "ETH",
 }
 
-export enum CredentialStatusType {
```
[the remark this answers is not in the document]
Alexander-frenki: in a case when we receive this data from BE I don't think that we need to have this enum, and when something will be added on BE we will not need to update this type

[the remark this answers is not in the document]
amonsosanz: I don't agree. We always need to ensure that the data we are managing is correct and fail fast and loud otherwise. Never trust the backend or any other sources. Credential status can only be [a subset of literals](https://github.com/0xPolygonID/js-sdk/blob/main/src/verifiable/constants.ts#L54-L68):

* `SparseMerkleTreeProof`
* `Iden3ReverseSparseMerkleTreeProof`
* `Iden3commRevocationStatusV1.0`
* `Iden3OnchainSparseMerkleTreeProof202`

Not any string is a valid `CredentialStatusType` and our types must reflect that reality.

> and when something will be added on BE we will not need to update this type

On the contrary! It is good that we are forced to update our code whenever a new member is added to the `CredentialStatusType` enum.

In reality, the right thing to do would be to import this type from the JS-SDK library, but since we don't have it as a dependency, we were cloning the type locally. At some point, we will add the dependency and remove this local type to import it from the library.

ui/src/adapters/api/identities.ts
```diff
@@ -160,20 +152,31 @@ export async function updateIdentityDisplayName({
   }
 }
 
-export const supportedNetworkParser = getStrictParser<SupportedNetwork>()(
-  z.object({
-    blockchain: z.string(),
-    networks: z.tuple([z.string()]).rest(z.string()),
-  })
+export const networkParser = z.object({
```
amonsosanz, line 155: `getStrictParser` is missing here

ui/src/adapters/api/identities.ts
```diff
@@ -160,20 +161,33 @@ export async function updateIdentityDisplayName({
...
+    rhsMode: z.tuple([z.nativeEnum(CredentialStatusType)]).rest(z.nativeEnum(CredentialStatusType)),
   })
 );
 
-export async function getSupportedNetwork({
+type BlockchainInput = Omit<Blockchain, "name"> & {
+  blockchain: string;
+};
+
+export const blockchainParser = getStrictParser<BlockchainInput, Blockchain>()(
```
amonsosanz, line 175: We don't need to export these parsers, right?

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -24,6 +22,34 @@ const initialValues: IdentityFormData = {
   type: IdentityType.BJJ,
 };
 
+function getNetworkFormValues(
+  blockchains: Blockchain[],
+  blockchainName?: string,
+  networkName?: string
+): Partial<IdentityFormData> | null {
+  const foundNetwork = blockchains.find(({ name }) => name === blockchainName);
```
amonsosanz, line 30: foundBlockchain

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -24,6 +22,34 @@ const initialValues: IdentityFormData = {
...
+  const selectedNetwork = foundNetwork ?? blockchains[0];
```
amonsosanz, line 31: selectedBlockchain

ui/src/components/identities/IdentityForm.tsx
```diff
@@ -24,6 +22,34 @@ const initialValues: IdentityFormData = {
...
+
+  if (!selectedNetwork) {
+    return null;
+  }
+
+  const { name, networks } = selectedNetwork;
+  const network = networks.find(({ name }) => name === networkName) || networks[0];
+  const parsedCredentialStatusType = credentialStatusTypeParser.safeParse(network.rhsMode[0]);
```
amonsosanz, line 39: We don't need to parse this.

`network.rhsMode[0]` is already of type `CredentialStatusType `

amonsosanz (approved):

martinsaporiti (approved):

PR #839, opened 2024-11-04
fix: docker-compose-full

martinsaporiti:


x1m3:
I think PR should go to develop branch.

x1m3 (approved):

infrastructure/local/docker-compose-full.yml
```diff
@@ -61,8 +61,10 @@ services:
       - "8088:80"
     env_file:
       - ../../.env-ui
+      - ../../.env-issuer
```
barrybecker4, line 64: what about ../ui/.env? Does that need to be loaded too? When I follow the README, this is what I see:
image

martinsaporiti: The `./ui/.env` file is needed only when you run the fe as a developer, executing `npm run ...`. It is not needed when you run the issuer node with docker. 
Regarding your problem, which steps are you following to run the issuer node?

barrybecker4: Ok. Never mind. I did finally get the issuer node running locally yesterday, but it was difficult to figure our the nginx and ngrok configuration needed to get past the CORS errors.

PR #841, opened 2024-11-04
chore: add links pagination

martinsaporiti:


internal/repositories/link.go
```diff
@@ -137,39 +137,73 @@ GROUP BY links.id, schemas.id
...
+		"schemas.id as schema_id",
+		"schemas.issuer_id as schema_issuer_id",
+		"schemas.url",
+		"schemas.type",
+		"schemas.hash",
+		"schemas.words",
+		"schemas.created_at",
+	}
+
+	//sql := `
```
x1m3, line 166: Remove this comment

x1m3 (commented):
Should we add UI frontend changes to same PR? I think it is a good idea.

martinsaporiti:
> Should we add UI frontend changes to same PR? I think it is a good idea.

yes, that is correct

PR #854, opened 2024-11-29
Add Keys functionality 

martinsaporiti:


internal/core/domain/claim.go
```diff
@@ -241,3 +241,12 @@ func (c *Claim) GetCredentialStatus() (*verifiable.CredentialStatus, error) {
 	}
 	return cStatus, nil
 }
+
+// GetPublicKey returns the public key of the claim
+// If the schema is not supported, it returns an unSupportedPublicKeyType
+func (c *Claim) GetPublicKey() PublicKey {
```
demonsh, line 247: @martinsaporiti I still insist to move this function to utilities, as you should have too much context to call it properly for one specific claim, this is knowledge that doesn't exist on the Claim level

internal/core/domain/publicKey.go
```diff
@@ -0,0 +1,50 @@
+package domain
+
+import (
+	"bytes"
+
+	"github.com/iden3/go-iden3-crypto/babyjub"
+)
+
+const (
+	bjjAuthSchemaURL = "https://schema.iden3.io/core/json/auth.json"
```
demonsh, line 10: Use constant from library.
I'll add to https://github.com/iden3/go-schema-processor/blob/main/verifiable/constants.go

demonsh (approved):

PR #866, opened 2024-12-18
fix: schema eth, country code validation

Alexander-frenki:


ui/src/components/credentials/IssueCredentialForm.tsx
```diff
@@ -144,6 +144,11 @@ export function IssueCredentialForm({
     }
   };
 
+  const isCountryCode = (x: number) => {
+    const iso31661NumericRegex = /^\d{1,3}$/;
+    return iso31661NumericRegex.test(x.toString());
```
amonsosanz, line 149: This will validate any 3-digit number, not valid country codes.

Alexander-frenki: I know, but we decided that this is  enough in this case, and using a list of countries will be redundant now

demonsh: Hi @amonsosanz, if you have more advanced validator please share it. We will added it to the node. 
But at the moment to allow testing, we will add just minor validation of 3 digits.

demonsh: We will add it as separate pull reqeust

amonsosanz: This is how we validate it in the schema builder:

https://github.com/0xPolygonID/schema-builder/pull/243/files#diff-2c90d10e1dc2218567877731655acf2af77eec2b038fbf5233a6806aee7d20b9

```
    const isCountryCode = (code: number) => {
      return iso31661Countries.find((country) => country.code === code) !== undefined;
    };
```

PR #870, opened 2024-12-23
add support of discover protocol

volodymyr-basiuk:
Added support for the `https://didcomm.org/discover-features/2.0/queries` protocol message:

- Supports the `accept`, `protocol`, and `header` feature types.
- Supports wildcards for queries (e.g., `match: 'https://iden3-communication.io/authorization/1.*'`).
- Extended `MediaTypeManager` with a function that returns supported protocol messages: `GetSupportedProtocolMessages` (this corresponds to the list of protocol feature types).
- Added unit-tests for /agent endpoint with discover query messages


internal/core/ports/discovery_service.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/polygonid/sh-id-platform/internal/core/domain"
+)
+
+// DiscoveryService is the interface implemented by the discovery service
+type DiscoveryService interface {
+	Agent(ctx context.Context, req *AgentRequest) (*domain.Agent, error)
+}
+
+// NewDiscoveryAgentRequest validates the inputs and returns a new AgentRequest
+func NewDiscoveryAgentRequest(basicMessage *comm.BasicMessage) (*AgentRequest, error) {
```
x1m3, line 20: This function is very similar to ports.NewAgentRequest.. I think we should try to unify. 
In a fast review, the only differences  I've found is the check for message type and the use of basicMessageID as claimID ¿!!!? But I think that this AgentRequest.ClaimID is not used so I guess it can be removed

internal/core/services/discovery.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, fmt.Errorf("invalid discover feature queries request body: %w", err)
+	}
+
+	disclosures := []protocol.DiscoverFeatureDisclosure{}
+	for _, query := range queries.Queries {
+		var disclosuresToAppend []protocol.DiscoverFeatureDisclosure
+		switch query.FeatureType {
+		case protocol.DiscoveryProtocolFeatureTypeAccept:
+			disclosuresToAppend, err = c.handleAccept(ctx)
+			if err != nil {
```
x1m3, line 56: log error

internal/core/services/discovery.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return disclosures
+}
+
+func wildcardToRegExp(match string) *regexp.Regexp {
+	// Escape special regex characters and replace `*` with `.*`
+	regexPattern := regexp.QuoteMeta(match)
+	regexPattern = strings.ReplaceAll(regexPattern, "\\*", ".*")
+	regExp, _ := regexp.Compile("^" + regexPattern + "$")
```
x1m3, line 149: Is it safe to ignore this error? Are we sure that regexPattern is valid?

x1m3 (commented):
LGTM. 

I left some comments. I think the more relevant is triying to unify new function NewDiscoveryAgentRequest with existing one.

internal/api/agent.go
```diff
@@ -20,16 +23,33 @@ func (s *Server) Agent(ctx context.Context, request AgentRequestObject) (AgentRe
 		return Agent400JSONResponse{N400JSONResponse{"cannot proceed with the given request"}}, nil
 	}
 
-	req, err := ports.NewAgentRequest(basicMessage)
-	if err != nil {
-		log.Error(ctx, "agent parsing request", "err", err)
-		return Agent400JSONResponse{N400JSONResponse{err.Error()}}, nil
-	}
+	var agent *domain.Agent
```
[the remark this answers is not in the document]
volodymyr-basiuk: I guess you meant `iden3comm.BasicMessage`, replaced

vmidyllic (approved):

x1m3 (approved):
LGTM!

martinsaporiti (approved):
LGMT

PR #880, opened 2024-12-31
chore: add public path to endpoints

martinsaporiti:


api/api.yaml
```diff
@@ -906,6 +935,29 @@ paths:
         '500':
           $ref: '#/components/responses/500'
 
+  /public/v2/identities/{identifier}/credentials/revocation/status/{nonce}:
```
x1m3, line 938: public/v2 or v2/public ? v2/public sounds better to me

x1m3 (commented):
This is a major API change. New major release?

PR #892, opened 2025-01-13
fix: verify payment

volodymyr-basiuk:
- The IsPaymentDone contract call should pass the signer address (the one recovered from the signature) as the first parameter, not the Recipient address.
- Refactor the payment verification process to simplify it.

### DIFF
--- a/internal/core/services/payment.go
+++ b/internal/core/services/payment.go
@@ -261,7 +261,7 @@ func (p *payment) verifyPaymentOnBlockchain(
 	txIdProvided := txID != nil && *txID != ""
 
 	if txIdProvided {
-		status, err := handleTransaction(ctx, client, *txID)
+		status, err := handlePaymentTransaction(ctx, client, *txID)
 		if err != nil || status != ports.BlockchainPaymentStatusSuccess {
 			return status, err
 		}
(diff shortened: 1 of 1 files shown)


demonsh (approved):

internal/core/services/payment.go
```diff
@@ -256,58 +254,58 @@ func (p *payment) verifyPaymentOnBlockchain(
...
+		return ports.BlockchainPaymentStatusSuccess, nil
+	}
+
 	return ports.BlockchainPaymentStatusFailed, nil
 }
 
-// paymentOptionConfigItem finds the payment option config item used to pay using the payment id stored in PaymentRequest database
-func (p *payment) paymentOptionConfigItem(ctx context.Context, issuerDID w3c.DID, item *domain.PaymentRequestItem) (*domain.PaymentOptionConfigItem, error) {
-	paymentReq, err := p.paymentsStore.GetPaymentRequestByID(ctx, issuerDID, item.PaymentRequestID)
+func handleTransaction(
```
x1m3, line 282: I would rename this function to something like handlePaymentTransaction or handleSomethingTransaction to avoid future collisions with other similar functions in package. 

But we could postpone this renaming while we don't have this problem,

x1m3 (approved):
LGTM. 

I jut left a small suggestion about a name but feel free to ignore it

internal/core/services/payment.go
volodymyr-basiuk: Sure, renamed

martinsaporiti (approved):

PR #904, opened 2025-01-22
refactor payment-rails signature generation

volodymyr-basiuk:


internal/core/services/payment.go
```diff
@@ -440,37 +442,19 @@ func (p *payment) paymentRequestSignature(
 		ID:   string(decodedKeyID),
 	}
 
+	types := apitypes.Types{}
+	fileBytes, err := os.ReadFile(fmt.Sprintf("schemas/%s.json", paymentType))
```
x1m3, line 446: This should be in a constructor (on initialization)  to avoid opening and parsing same file on each execution.

internal/api/schemas/Iden3PaymentRailsRequestV1.json
```diff
@@ -0,0 +1,42 @@
+{
```
x1m3, line 1: As this information doesn't seems configurable, I think it is better to maintain it in a constant in code and not in a file.

x1m3 (approved):
LGTM. Less code than before ;-9

internal/core/services/payment.go
```diff
@@ -33,21 +34,29 @@ import (
...
+	kms                                  kms.KMSType
+	iden3PaymentRailsRequestV1Types      apitypes.Types
+	iden3PaymentRailsERC20RequestV1Types apitypes.Types
 }
 
 // NewPaymentService creates a new payment service
 func NewPaymentService(payOptsRepo ports.PaymentRepository, resolver network.Resolver, schemaSrv ports.SchemaService, settings *payments.Config, kms kms.KMSType) ports.PaymentService {
+	iden3PaymentRailsRequestV1Types := apitypes.Types{}
+	iden3PaymentRailsERC20RequestV1Types := apitypes.Types{}
+	_ = json.Unmarshal([]byte(domain.Iden3PaymentRailsRequestV1SchemaJSON), &iden3PaymentRailsRequestV1Types)
```
vmidyllic, line 50: is it ok to ommit error?

volodymyr-basiuk: I think yes if we work with `const` variables.
@x1m3  what do you think ?

vmidyllic (approved):

martinsaporiti (approved):

PR #909, opened 2025-01-27
feat: payment requests status

volodymyr-basiuk:
- added `modifiet_at`, `status`, `paid_nonce` to `payment_requests` table
- update these fields on payment verification
- if status is `success`, than `paid_nonce` is not empty


martinsaporiti (approved):
LGTM

x1m3 (approved):
GOOD!

PR #938, opened 2025-10-13
chore: add encryption key

martinsaporiti:


api/api.yaml
```diff
@@ -2738,6 +2785,14 @@ components:
           example: "did:iden3:polygon:amoy:2qVSDLNYs2s5HDMTyb4Su67ZvrZ..."
           description: If a User DID is provided, the payment will be verified for the specified User DID.
 
+    JWE:
```
ilya-korotya: Do we use it some where? Can we remove it?

internal/api/credentials.go
```diff
@@ -100,6 +101,10 @@ func (s *Server) CreateCredential(ctx context.Context, request CreateCredentialR
 	req := ports.NewCreateClaimRequest(did, request.Body.ClaimID, request.Body.CredentialSchema, request.Body.CredentialSubject, expiration, request.Body.Type, request.Body.Version, request.Body.SubjectPosition, request.Body.MerklizedRootPosition, claimRequestProofs, nil, false, *credentialStatusType, toVerifiableRefreshService(request.Body.RefreshService), request.Body.RevNonce,
 		toVerifiableDisplayMethod(request.Body.DisplayMethod))
 
+	if request.Body.EncryptionKey != nil {
+		req.WithEncryptionKey(*request.Body.EncryptionKey)
```
ilya-korotya: Why do we need this separate method? If we can pass `request.Body.EncryptionKey` directly to `NewCreateClaimRequest` constructor?

internal/core/ports/claim_service.go
```diff
@@ -86,36 +89,10 @@ type ClaimsFilter struct {
...
-	if queryField != nil {
-		filter.QueryField = *queryField
-	}
-	if queryValue != nil {
-		filter.QueryFieldValue = *queryValue
-	}
-
-	return &filter, nil
+// WithEncryptionKey sets the encryption key for the claim request
+func (cr *CreateClaimRequest) WithEncryptionKey(encryptionKey EncryptionKey) *CreateClaimRequest {
```
ilya-korotya: Why we can't use constructor?

internal/core/domain/claim.go
```diff
@@ -246,3 +250,32 @@ func (c *Claim) GetCredentialStatus() (*verifiable.CredentialStatus, error) {
...
 }
+
+// HasEncryptedData returns true if the claim has encrypted data
+func (c *Claim) HasEncryptedData() bool {
+	return c.Data.Status == pgtype.Null && c.EncryptedData != nil
+}
+
+// GetVerifiableProofs returns the verifiable proofs of the claim
+func (c *Claim) GetVerifiableProofs() (verifiable.CredentialProofs, error) {
+	var err error
```
ilya-korotya: use var block to define variables:
```go
var (
    err error
    proofs ...
)
```

internal/core/domain/claim.go
```diff
@@ -246,3 +250,32 @@ func (c *Claim) GetCredentialStatus() (*verifiable.CredentialStatus, error) {
...
+	var signatureProof *verifiable.BJJSignatureProof2021
+	if c.SignatureProof.Status != pgtype.Null {
+		err = c.SignatureProof.AssignTo(&signatureProof)
+		if err != nil {
+			return nil, err
+		}
+		proofs = append(proofs, signatureProof)
+	}
+
+	var mtpProof *verifiable.Iden3SparseMerkleTreeProof
```
ilya-korotya: You can move the var inside to if statement.

internal/core/domain/claim.go
```diff
@@ -246,3 +250,32 @@ func (c *Claim) GetCredentialStatus() (*verifiable.CredentialStatus, error) {
...
+// HasEncryptedData returns true if the claim has encrypted data
+func (c *Claim) HasEncryptedData() bool {
+	return c.Data.Status == pgtype.Null && c.EncryptedData != nil
+}
+
+// GetVerifiableProofs returns the verifiable proofs of the claim
+func (c *Claim) GetVerifiableProofs() (verifiable.CredentialProofs, error) {
+	var err error
+	proofs := make(verifiable.CredentialProofs, 0)
+	var signatureProof *verifiable.BJJSignatureProof2021
```
ilya-korotya: You can move the var inside to if statement.

internal/core/services/claims.go
```diff
@@ -203,6 +200,11 @@ func (c *claim) CreateCredential(ctx context.Context, req *ports.CreateClaimRequ
 		opts.MerklizerOpts = []merklize.MerklizeOption{merklize.WithDocumentLoader(c.loader)}
 	}
 
+	vc, err := c.createVC(ctx, req, vcID, jsonLdContext, nonce)
```
ilya-korotya, line 205: What is the reason move the `c.createVC` from line 179 to this line 203?

internal/core/services/claims.go
```diff
@@ -283,6 +257,78 @@ func (c *claim) CreateCredential(ctx context.Context, req *ports.CreateClaimRequ
...
+
+// setVC sets the verifiable credential or the encrypted data in the claim.
+// If the encryption key is provided, it encrypts the verifiable credential and sets the encrypted data in the claim.
+// Otherwise, it sets the verifiable credential in the claim.
+// The encryption is done using the AnoncryptPacker from iden3comm.
+// The claim parameter is modified in place.!
+func setVC(ctx context.Context, req *ports.CreateClaimRequest, jsonLdContext string, vc verifiable.W3CCredential, claim *domain.Claim) error {
+	if req.EncryptionKey != nil {
+		claim.ContextUrl = &jsonLdContext
+		return encryptCredential(ctx, req, vc, claim)
```
ilya-korotya: refactor to `claim.EncryptedData = encryptCredential(ctx, req, vc)`
Don't pass claim as parameter only for update an value.

internal/core/services/claims.go
```diff
@@ -823,17 +868,27 @@ func (c *claim) getAgentCredential(ctx context.Context, basicMessage *ports.Agen
 		return nil, err
 	}
 
-	vc, err := schemaPkg.FromClaimModelToW3CCredential(*claim)
-	if err != nil {
-		log.Error(ctx, "creating W3 credential", "err", err)
-		return nil, fmt.Errorf("failed to convert claim to  w3cCredential: %w", err)
-	}
+	var body []byte
+	if !claim.HasEncryptedData() {
```
ilya-korotya: Maybe is better to remove `!` to simplify the condition?
```go
if claim.HasEncryptedData() { create encrypted response message }
else { create non encrypted response message }
```

api/api.yaml
```diff
@@ -2599,14 +2615,42 @@ components:
...
+        - data
+        - proof
+        - context
+        - type
+      properties:
+        id:
+          type: string
+          x-omitempty: false
+        data:
+          type: string
```
ilya-korotya: data is object not string

ilya-korotya (approved):

internal/api/credentials.go
martinsaporiti: We can send it in the constructor, of course.  For a large number of fields in the constructor (not required). I prefer this.

internal/core/services/claims.go
martinsaporiti: because we were creating the VC before loading the schema....

internal/core/services/claims.go
martinsaporiti: We are not passing the claim; we are passing the pointer.

ilya-korotya: I mean you pass pointer to claim, to set a value to claim inside the `encryptCredential` function.
On my point of view `claim.EncryptedData = encryptCredential(ctx, req, vc)` is more clear.

api/api.yaml
```diff
@@ -2599,14 +2615,42 @@ components:
...
+          example:  "KYCAgeCredential"
+
     Credential:
       type: object
       required:
         - id
         - proofTypes
         - revoked
         - schemaHash
-        - vc
```
ilya-korotya, line 2609: Move this field to the optional section instead of delete it

internal/core/domain/claim.go
```diff
@@ -246,3 +251,50 @@ func (c *Claim) GetCredentialStatus() (*verifiable.CredentialStatus, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		proofs = append(proofs, mtpProof)
+	}
+	return proofs, nil
+}
+
+// GetEncryptedDataAsMap returns the encrypted data as a map
+func (c *Claim) GetEncryptedDataAsMap() (map[string]any, error) {
```
ilya-korotya: We have separate type on the verifiable package to represent JWE token: https://github.com/iden3/iden3comm/blob/main/protocol/credentials.go#L515

ilya-korotya (approved):

martinsaporiti (commented):
LGTM

PR #941, opened 2025-10-15
chore: add tests

martinsaporiti:


internal/core/services/claims_test.go
```diff
@@ -344,3 +344,23 @@ func TestBuildEncryptedCredentialBody(t *testing.T) {
...
+		Network:    net,
+		Method:     method,
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, identity)
+	identifier := identity.Identifier
+	did, err := w3c.ParseDID(identifier)
+	assert.NoError(t, err)
+
+	t.Run("should return the first non-revoked auth claim", func(t *testing.T) {
```
ilya-korotya: This test doesn't test `should return the first non-revoked auth claim` since we created only one auth claim.

ilya-korotya (commented):
Consider using `require` instead of `assert`. In your case, there is no point in continuing testing if an error has occurred.

PR #942, opened 2025-10-17
fix: schema type

martinsaporiti:


x1m3 (approved):

internal/repositories/claim.go
```diff
@@ -823,8 +823,8 @@ func processClaims(rows pgx.Rows) ([]*domain.Claim, error) {
 		err := rows.Scan(&claim.ID,
 			&claim.Issuer,
 			&claim.SchemaHash,
-			&claim.SchemaType,
 			&claim.SchemaURL,
+			&claim.SchemaType,
```
ilya-korotya, line 828: The method `GetAuthClaimsForPublishing` scans the fields in another siquence:
```
       	schema_hash,
       	schema_type,
       	schema_url,
```

ilya-korotya (changes requested):

api/api.yaml
```diff
@@ -2615,6 +2615,24 @@ components:
         connection:
           $ref: '#/components/schemas/AuthenticationConnection'
 
+
+    CredentialStatus:
```
ilya-korotya: You can create reference to type `verifiable.CredentialStatus` and then reuse it in `required` section and `properties` section.

martinsaporiti: yes, I did it in that way but I forgot to remove the type from api spec

ilya-korotya (approved):

PR #948, opened 2025-11-27
feat(cache) TLS config

Strnadj:


x1m3:
Hi @Strnadj , and thank you very much for this PR!

We really appreciate you taking the time to contribute. According to our contribution guidelines (contrib.md), all pull requests should target the devel branch instead of main, so that we can keep main stable.

Could you please reopen this PR (or open a new one) against the devel branch?

x1m3:
Regarding your PR, have you tried to pass redis url starting with rediss like "rediss://:password@host:6379/0" ? It should work and it is the simple way to configure a secure connection without adding more config params

Strnadj:
> Regarding your PR, have you tried to pass redis url starting with rediss like "rediss://:password@host:6379/0" ? It should work and it is the simple way to configure a secure connection without adding more config params

For redis maybe, (doc says something else https://redis.uptrace.dev/guide/go-redis.html#using-tls) I can check that later, but valkey has no option like this...

Strnadj:
@x1m3 I was able to test rediss and it's working... so I will remove that.

Related base branch it has been changed to devel.

